PR #82, opened 2023-03-23
3D demo deployment

markhougaard:


ammapspeople:
@markhougaard I have a hunch that this is not needed anymore. Can we close it?

PR #88, opened 2023-03-27
feat: Make the map show MapsIndoors Live Data if available.

ammapspeople:
# What

- Show default MapsIndoors Live Data badges always (when available on the solution).

# How

- Implement wrapper for MapsIndoors LiveDataManager in a custom React hook.
- The hook will ask the backend for available Live Data Domain Types on the current solution, and enable those Domain Types on the map.
  - The logic is per best practice wrapped in an async function within a useEffect hook
- When solution changes, some cleanup (disabling) is done before re-enabling the Domain Types for the new solution.
- Added and used apiKey prop to Map component. It was already given.


andreeaceachir142 (approved):
👍

PR #98, opened 2023-03-29
Bugfix/categories selected venue

andreeaceachir142:
# What 
- Show categories only for the selected venue.

# How 
- Create a function `getVenueCategories` which filters the locations to only return the ones that have the same venue as the currently selected venue.


packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -97,12 +98,22 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         }
     }
 
+    /**
+     * Get the categories for the selected venue.
```
ammapspeople, line 102: Missing JSDoc param

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -97,12 +98,22 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
...
+     */
+    function getVenueCategories(venue) {
+        mapsindoors.services.LocationsService.getLocations({}).then(locations => {
```
ammapspeople, line 105: Write a comment about us not being able to get the venue parameter to work here, which is why we do the filtering. Also create a ticket in the SDK project about that, with steps to reproduce.

ammapspeople (changes requested):

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -97,12 +98,27 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         }
     }
 
+    /**
+     * Get the categories for the selected venue.
+     *
+     * @param {string} venue
+     */
+    function getVenueCategories(venue) {
+        console.log('venue', venue);
```
ammapspeople, line 107:
```suggestion
```

ammapspeople (approved):
👍

PR #99, opened 2023-03-29
Bugfix/venue selector visibility

andreeaceachir142:
# What 
- Hide the venue selector when having directions open 

# How 
- Refactor the `BottomSheet` and `Sidebar` components to keep track of the state of the directions page 
- Call the `onDirectionsOpened` and `onDirectionsClosed` when navigating to and from the directions page
- Refactor the implementation of toggling the visibility of the floor selector to be more generic, as we need to show/hide more than one element on the map
- Refactor the `MapsIndoorsMap` component to keep track of directions and hide or show the floor selector and the venue selector

### DIFF
--- a/packages/map-template/src/components/BottomSheet/BottomSheet.jsx
+++ b/packages/map-template/src/components/BottomSheet/BottomSheet.jsx
@@ -21,14 +21,12 @@ const BOTTOM_SHEETS = {
  * @param {Object} props.setCurrentLocation - The setter for the currently selected MapsIndoors Location.
  * @param {Object} props.currentCategories - The unique categories displayed based on the existing locations.
  * @param {function} props.onLocationsFiltered - The list of locations after filtering through the categories.
- * @param {function} props.onHideFloorSelector - Trigger the visibility of the floor selector to be hidden.
- * @param {function} props.onShowFloorSelector - Trigger the visibility of the floor selector to be shown.
  * @param {function} props.onDisableLocations - Restrict the user from interacting with the locations when in directions mode.
  * @param {function} props.onEnableLocations - Allow the user to interact with the locations when outside of directions mode.
- * @param {function} props.onHideVenueSelector - Trigger the visibility of the venue selector to be hidden.
- * @param {function} props.onShowVenueSelector - Trigger the visibility of the venue selector to be shown.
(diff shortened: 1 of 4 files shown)


packages/map-template/src/components/BottomSheet/BottomSheet.jsx
```diff
@@ -21,12 +21,12 @@ const BOTTOM_SHEETS = {
...
  * @param {function} props.onLocationsFiltered - The list of locations after filtering through the categories.
- * @param {function} props.onHideFloorSelector - Trigger the visibility of the floor selector to be hidden.
- * @param {function} props.onShowFloorSelector - Trigger the visibility of the floor selector to be shown.
  * @param {function} props.onDisableLocations - Restrict the user from interacting with the locations when in directions mode.
  * @param {function} props.onEnableLocations - Allow the user to interact with the locations when outside of directions mode.
+ * @param {function} props.onDirectionsOpened - Check if the directions page state is open.
+ * @param {function} props.onDirectionsClosed - Check if the directions page state is closed.
  */
-function BottomSheet({ currentLocation, setCurrentLocation, currentCategories, onLocationsFiltered, onHideFloorSelector, onShowFloorSelector, onDisableLocations, onEnableLocations}) {
+function BottomSheet({ currentLocation, setCurrentLocation, currentCategories, onLocationsFiltered, onDisableLocations, onEnableLocations, onDirectionsOpened, onDirectionsClosed}) {
```
ammapspeople, line 29: Do we need both the pairs for locations and directionsOpened? Can they be combined into one?

andreeaceachir142: For the disable/enable locations?

ammapspeople: I was wondering if we need all four:

`onDisableLocations, onEnableLocations, onDirectionsOpened, onDirectionsClosed}`

Are they (pairwise) handling the same thing? Not sure.

andreeaceachir142: That is true actually, the enable/disable location is something that is dependent on the state of directions, so I refactored the code and removed the unnecessary parameters.

ammapspeople (approved):
👏

PR #101, opened 2023-03-29
Bugfix/wayfinding error message

andreeaceachir142:
# What 
- Reset the state of the error messages when clearing or searching again.

# How
- Change the state of the error messages shown when `onSearchCleared` and `onSearchClicked` are called.

### DIFF
--- a/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
+++ b/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
@@ -46,6 +46,7 @@ function Wayfinding({ onStartDirections, onBack, location, onSetSize, isActive, 
     /** Check if search results have been found */
     const [hasSearchResults, setHasSearchResults] = useState(true);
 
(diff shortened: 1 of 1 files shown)


packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -46,6 +46,7 @@ function Wayfinding({ onStartDirections, onBack, location, onSetSize, isActive,
     /** Check if search results have been found */
     const [hasSearchResults, setHasSearchResults] = useState(true);
 
+    /** Check if the searched route throws errors */
```
ammapspeople, line 49: I don't think this useState is a "check", but more an "indicator".

The same applies to the one above.

andreeaceachir142: ✅

ammapspeople (approved):
✅

PR #105, opened 2023-04-03
Feature/App user roles

andreeaceachir142:
# What 
- Add support for app user roles array 

# How 
- Add a new property on the `MapsIndoorsMap` component for the `appUserRoles`
- React to changes on the app user roles property
- Use the SDK functions to get the app user roles and filter through them to get the ones that match the name 
- Set the app user roles with the ones that match


ammapspeople (approved):
👏

PR #106, opened 2023-04-03
Use web component wrapper components for search & categories

ammapspeople:
# What

- Use web component wrappers for searching and category handling.

# How

- Refactored `WebComponentWrappers/Search` to utilize `useImperativeHandle` in order to programmatically trigger methods on the `<mi-search>` component without relying on props.
- Introduced category prop on `WebComponentWrappers/Search` so we can filter by a category.
- Refactored the `Search` component to use the `WebComponentWrappers/Search` instead of `<mi-search>` directly.
  - This had the benefits of
    - removing some usages of `document.createElement` and native DOM event listeners
    - getting rid of "double" variable for holding the selected category.
- Refactored the `Wayfinding` component to make use of the imperative ways to trigger method on the `WebComponentWrappers/Search` component.

### DIFF
--- a/packages/map-template/src/components/Search/Search.jsx
+++ b/packages/map-template/src/components/Search/Search.jsx
@@ -28,11 +28,10 @@ let _selectedCategory;
  */
 function Search({ onLocationClick, categories, onLocationsFiltered, onSetSize }) {
 
-    /** Referencing the search DOM element */
+    /** Referencing the search field */
     const searchFieldRef = useRef();
 
     const [searchResults, setSearchResults] = useState([]);
-    const [searchFieldValue, setSearchFieldValue] = useState();
 
     /** Referencing the search results container DOM element */
     const searchResultsRef = useRef();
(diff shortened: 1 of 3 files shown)


andreeaceachir142:
Having the wrapper component for the search, could we get rid of the private property `_activeSearchField` on line 24 on the `Wayfinding` page?

andreeaceachir142:
When searching for a location i.e. `Dining room`, then going to the Location details page and then to the `Wayfinding` page, we have the `FROM` field on focus.
If I try clicking on the `TO` field **before** typing anything in the `FROM` field, then going back to the `FROM` field and selecting another location i.e. `Reception`, then I do not get any information about the accessibility, distance and time. 🤔

ammapspeople:
> Having the wrapper component for the search, could we get rid of the private property `_activeSearchField` on line 24 on the `Wayfinding` page?

I'm not sure it can be fixed based on these changes, but I will give it a try.

ammapspeople:
> When searching for a location i.e. `Dining room`, then going to the Location details page and then to the `Wayfinding` page, we have the `FROM` field on focus. If I try clicking on the `TO` field **before** typing anything in the `FROM` field, then going back to the `FROM` field and selecting another location i.e. `Reception`, then I do not get any information about the accessibility, distance and time. 🤔

Thanks, I'll look into that.

ammapspeople:
> Thanks, I'll look into that.

I can reproduce the same in main and on our currently deployed demo.
It's not related to these changes. I'll create a separate ticket to fix it.

ammapspeople:
> Having the wrapper component for the search, could we get rid of the private property `_activeSearchField` on line 24 on the `Wayfinding` page?

This is now fixed.

andreeaceachir142 (approved):

PR #107, opened 2023-04-04
Bugfix/search results scroll

andreeaceachir142:
# What 
- Fix the search results not being scrollable when navigating back to the Search page from the Location Details page

# How 
- Set the style of the `sheet__content` element to follow the same style of the `sheet` element


packages/map-template/src/components/BottomSheet/Sheet/Sheet.jsx
```diff
@@ -65,6 +65,7 @@ function Sheet({ children, isOpen, minHeight, preferredSizeSnapPoint, onSwipedTo
             switch (targetSize) {
                 case snapPoints.MAX:
                     setStyle({ height: `${container.current.clientHeight}px`});
+                    console.log(container.current);
```
ammapspeople, line 68: 🖐️ 😄

andreeaceachir142: 🫣

ammapspeople (approved):

PR #108, opened 2023-04-04
Fix/avoid replacing imageurl on location.details

ammapspeople:
# What

- Fix bug where imageURL on a Locations was wrongly populated with icon from Display Rule when having a Location displayed as part of a search result.

# How

- Implement new (optional) icon prop on the Stencil component `<mi-list-item-location>`
  - The component will fallback to current behavior if not set, thus ensuring backwards comparability.
  - This required some refactoring of the Live Data badge handling
- Set the icon prop within the wrapper component `ListItemLocation` instead of overwriting the Location data in the `Search` component.

### DIFF
--- a/packages/components/src/components/list-item-location/list-item-location.tsx
+++ b/packages/components/src/components/list-item-location/list-item-location.tsx
@@ -1,5 +1,5 @@
 import midtIcon from '@mapsindoors/midt/tokens/icon.json';
-import { Component, Event, EventEmitter, h, Host, JSX, Prop, Watch } from '@stencil/core';
+import { Component, Event, EventEmitter, h, Host, JSX, Prop, State, Watch } from '@stencil/core';
 import { appendMapsIndoorsImageQueryParameters } from '../../utils/utils';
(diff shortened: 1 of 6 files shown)


packages/components/src/components/list-item-location/list-item-location.tsx
```diff
@@ -57,6 +63,8 @@ export class ListItemLocation {
      */
     @Event() listItemDidRender: EventEmitter;
 
+    @State() iconURLToRender; string;
```
andreeaceachir142, line 66: What is the purpose of using the `State`?

ammapspeople: It makes sure the component is re-rendered when it changes, in contrast to a private variable, which would do no such thing.

andreeaceachir142: Alright, good to know. Thanks

andreeaceachir142 (approved):

PR #109, opened 2023-04-05
Feature/directions all locations

andreeaceachir142:
# What 
- Reset the filtered locations on the map when in Directions mode

# How 
- Create a state for the initial filtered locations and assign the filtered locations, in order to be able to reset the locations and reassign the filters when navigating out of Direction

### DIFF
--- a/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
+++ b/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
@@ -38,7 +38,7 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
     const [currentVenueName, setCurrentVenueName] = useState();
     const [currentLocation, setCurrentLocation] = useState();
     const [currentCategories, setCurrentCategories] = useState([]);
-    const [filtereLocations, setFilteredLocations] = useState();
+    const [filteredLocations, setFilteredLocations] = useState();
     const [initialFilteredLocations, setInitialFilteredLocations] = useState();
     const [mapsIndoorsInstance, setMapsIndoorsInstance] = useState();
     const [directionsService, setDirectionsService] = useState();
@@ -191,7 +191,7 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
      */
     useEffect(() => {
         if (hasDirectionsOpen) {
-            setInitialFilteredLocations(filtereLocations)
+            setInitialFilteredLocations(filteredLocations)
             setFilteredLocations([]);
         } else {
             setFilteredLocations(initialFilteredLocations);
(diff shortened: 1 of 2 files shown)


packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -161,7 +162,7 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         }
     }, [locationId]);
 
-    /*
+    /**
```
ammapspeople, line 165: I realize now that this is just a function call, and not a function definition. Thus is should not be a JSDoc block but just a regular comment. I believe we've used JSDoc blocks a lot of other places too where it wasn't relevant.

```suggestion
    /*
```

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -39,6 +39,7 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
     const [currentLocation, setCurrentLocation] = useState();
     const [currentCategories, setCurrentCategories] = useState([]);
     const [filteredLocations, setFilteredLocations] = useState();
+    const [initialFilteredLocations, setInitialFilteredLocations] = useState();
```
ammapspeople, line 42: Write some comments so it will be easier to differentiate between filteredLocations and initialFilteredLocations.

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -184,6 +185,22 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
...
+     * React on changes in directions opened state.
+     */
+    useEffect(() => {
+        // Reset all the filters when in directions mode.
+        // Store the filtered locations in another state, to be able to access them again.
+        if (hasDirectionsOpen) {
+            setInitialFilteredLocations(filteredLocations)
+            setFilteredLocations([]);
+        } else {
+        // Apply the previously filtered locations to the map when navigating outside the directions.
```
ammapspeople, line 199:
```suggestion
            // Apply the previously filtered locations to the map when navigating outside the directions.
```

ammapspeople (changes requested):

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
andreeaceachir142: ✅

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -184,6 +189,22 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         });
     }, [apiKey]);
 
+
+    /**
```
ammapspeople, line 193: Also not a function definition 😄 
```suggestion
    /*
```

ammapspeople (changes requested):

ammapspeople (approved):

PR #110, opened 2023-04-05
Feature/mimic browser history

ammapspeople:
# What
- Make the browser history pick up changes to significant app state changes.

# Why
- We want the browser back/forward button to work when changing state in the app.

# How
- Introduce `useAppHistory` hook that manages an app state and uses the browser history API.
- Refactored the VenueSelector, BottomSheet and the Modal to use and set the app state instead of it's own active sheet or modal and state.


andreeaceachir142 (approved):
👏

PR #111, opened 2023-04-05
Bugfix/wayfinding navigation

andreeaceachir142:
# What 
- Fix the inability to get directions in the wayfinding page 

# How 
- Refactor the handling of the clicked search fields
- Create a new function which programatically triggers the search when having a value

### DIFF
--- a/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
+++ b/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
@@ -33,7 +33,6 @@ function Wayfinding({ onStartDirections, onBack, location, onSetSize, isActive, 
     const detailsRef = useRef();
 
     const toFieldRef = useRef();
-
     const fromFieldRef = useRef();
 
     const directionsService = useContext(DirectionsServiceContext);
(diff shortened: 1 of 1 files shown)


packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -134,25 +140,45 @@ function Wayfinding({ onStartDirections, onBack, location, onSetSize, isActive,
      */
     function onSearchCleared(searchFieldIdentifier) {
         setActiveSearchField(searchFieldIdentifier);
```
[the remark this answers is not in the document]
andreeaceachir142: Yes, we use it on line 74 on the function `locationClickHandler` to check the active search field based on the identifier.

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -48,13 +49,18 @@ function Wayfinding({ onStartDirections, onBack, location, onSetSize, isActive,
...
     const [hasError, setHasError] = useState(false);
 
+     /** Indicate if the search has been triggered */
+    const [searchTriggered, setSearchTriggered] = useState(false);
+
     /** Holds search results given from a search field */
     const [searchResults, setSearchResults] = useState([]);
 
     const [destinationLocation, setDestinationLocation] = useState();
+
```
ammapspeople, line 59: My personal preference is to group related states, refs etc. without empty lines between them. I find that to be more readable.

andreeaceachir142: It makes sense, let's stick to this pattern 👍

PR #112, opened 2023-04-05
Feature/icon buttons

andreeaceachir142:
# What 
- Remove icons from wayfinding and directions buttons based on UX spec.

# How 
- Remove the icons and the corresponding styles.


ammapspeople (approved):
👍

PR #114, opened 2023-04-13
fix/wrong app size

ammapspeople:
# What, how

- Avoid using 100vh for app size.

# Why

- 100vh is more than visually visible in some mobile browsers.


andreeaceachir142 (approved):
👍

PR #115, opened 2023-04-13
Feature/map template styles

andreeaceachir142:
# What
- Align the map template styles based on https://mapspeople.atlassian.net/wiki/spaces/MIGEN/pages/3177480193/UI+corrections+checklist

# How 
- Adjust the CSS according to the latest design

### DIFF
--- a/packages/map-template/src/components/LocationDetails/LocationDetails.scss
+++ b/packages/map-template/src/components/LocationDetails/LocationDetails.scss
@@ -13,7 +13,7 @@
         &::before {
             content: '';
             position: absolute;
-            bottom: calc(56px + var(--spacing-medium));
+            bottom: calc(64px + var(--spacing-medium));
             left: 0;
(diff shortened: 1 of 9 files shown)


packages/map-template/src/components/LocationDetails/LocationDetails.scss
```diff
@@ -6,7 +6,7 @@
     display: grid;
     height: 100%;
     overflow-y: hidden;
-    grid-template-rows: 40px 1fr 56px;
+    grid-template-rows: 40px 1fr 64px;
```
ammapspeople, line 9: It seems the 56px value is also used on line 16. Should that be changed as well?

andreeaceachir142: ✅

ammapspeople (approved):

PR #116, opened 2023-04-13
Bugfix/reset categories

andreeaceachir142:
# What 
- Reset categories when changing the venue 

# How 
- Pass the `currentVenueName` prop to the `Sidebar` and `Bottom sheet`, and then to the `Search` component
- Implement `useEffect` hook and listen to the changes in the `currentVenueName`
- If there is a selected category, deselect it and clear the results list


ammapspeople:
Do these changes break anything in case no venue prop is provided to the map template?

andreeaceachir142:
No, because if there is no venue prop then we also don't have any categories to show, unless you select a venue from the venue selector.

ammapspeople:
> No, because if there is no venue prop then we also don't have any categories to show, unless you select a venue from the venue selector.

I am not sure that is the intended behaviour, since a venue should always be chosen. But it's definitely not related to those changes anyway.

ammapspeople (approved):
✅

PR #117, opened 2023-04-13
Feature/switch directions

andreeaceachir142:
# What 
- Implement switch between the origin and the destination locations. 

# How
- Implement a click handler for the switch button between the origin and destination locations.
- Add support for switching the fields when either of them is left empty.
- Add a new method on the search field wrapper component to programatically trigger the clear event on the input field.


ammapspeople (approved):
👏

packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
```diff
@@ -33,6 +33,9 @@ const SearchField = forwardRef(({ placeholder, mapsindoors, results, clicked, cl
         },
         focusInput() {
             elementRef.current.focusInput();
+        },
+        clearInput() {
```
ammapspeople, line 37: We could rename it to reflect the "original" method name on the <mi-search>, "clear".

andreeaceachir142: Done ✅

PR #119, opened 2023-04-13
Bugfix/reset location

andreeaceachir142:
# What 
- Fix the impossibility to click on the same location after navigating back to the search page 

# How
- Implement function for navigating back to the search page and reset the location, in order to be able to select it again


ammapspeople (approved):
👍

PR #120, opened 2023-04-17
Fix/avoid content pushing

ammapspeople:
# What

- Prevent mobile browsers to push input fields away from viewport when opening keyboard.

# How

- Overall strategy: Avoid applying focus while running a transition.
- Extended the `<mi-search>` component with a `disabled` attribute.
- On the Search component, use the `disabled` attribute to prevent the input field from gaining focus when clicking. Instead wait for the bottom sheet transition to finish before applying it.
- On the Wayfinding component, wait for bottom sheet transition to finish before applying focus.


andreeaceachir142 (approved):
👏

PR #122, opened 2023-04-18
Bugfix/list item location

andreeaceachir142:
# What 
- Fix the issue of not having a location and assigning the `imageURL`

# How
- Add a check for the existence of the location before assigning the `imageURL` to it


ammapspeople:
👍

PR #123, opened 2023-04-19
Feature/categories list icons

andreeaceachir142:
# What 
- Add icons from App Config to the categories list 

# How
- Add `iconUrl` when constructing the category info object in the `MapsIndoorsMap` component
- Get the categories from the App Config and find the ones with the matching key 
- Assign the icon url to the newly created key in the object `iconUrl`
- Use the `iconUrl` in the `Search` component when rendering the `mi-chip` component

### DIFF
--- a/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
+++ b/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
@@ -120,12 +120,16 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         // Build an object which contains the key, the count and the display name.
         for (const location of locationsResult) {
             const keys = Object.keys(location.properties.categories);
+
             for (const key of keys) {
+                const appConfigCategories = appConfigResult?.menuInfo.mainmenu;
+                const appConfigCategory = appConfigCategories.find(category => category.categoryKey === key);
+
                 if (uniqueCategories.has(key)) {
                     let count = uniqueCategories.get(key).count;
-                    uniqueCategories.set(key, { count: ++count, displayName: location.properties.categories[key] });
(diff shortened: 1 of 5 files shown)


packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -119,12 +120,17 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         // Build an object which contains the key, the count and the display name.
         for (const location of locationsResult) {
             const keys = Object.keys(location.properties.categories);
+
             for (const key of keys) {
+                // Get the categories from the App Config and find the ones with a matching key.
+                const appConfigCategories = appConfigResult?.menuInfo.mainmenu;
+                const appConfigCategory = appConfigCategories.find(category => category.categoryKey === key);
```
ammapspeople, line 127: I don't think it's necessary to have the `appConfigCategories` const. We could just use `appConfigResult?.menuInfo.mainmenu` on this line.

andreeaceachir142: I agree 👍

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -119,12 +120,16 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
         // Build an object which contains the key, the count and the display name.
```
ammapspeople, line 120:
```suggestion
        // Build an object which contains the key, the count, the display name and an icon.
```

ammapspeople (approved):
👏

PR #124, opened 2023-04-19
Props for initially showing wayfinding

ammapspeople:
# What

- Make it possible to set props to instantly show directions.

# How

- The MapsIndoorsMap component was extended with the `directionsFrom` and `directionsTo` props, both taking MapsIndoors Location IDs as strings. Those are converted to MapsIndoors Location objects and passed down to underlying components as props.
- The BottomSheet and Sidebar components now listens for changes directionsFromLocation and directionsToLocation and will show the Wayfinding component if both are set.
- - The Wayfinding component was refactored from having a `location` prop to have `from` and `to` props.


packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -199,14 +200,21 @@ function Wayfinding({ onStartDirections, onBack, location, onSetSize, isActive,
...
-        if (location) {
-            toFieldRef.current.setDisplayText(location.properties.name);
-            setDestinationLocation(location);
+        // If there is a "to" location, use that as the "to" field.
+        if (to) {
+            toFieldRef.current.setDisplayText(to.properties.name);
+            setDestinationLocation(to);
+        }
+
+        // If there is an "from" location, use that as the 'from' field. Otherwise trigger focus on search field.
```
andreeaceachir142, line 209:
```suggestion
        // If there is a "from" location, use that as the 'from' field. Otherwise trigger focus on search field.
```

PR #127, opened 2023-04-25
Bugfix/ switch directions

andreeaceachir142:
# What 
- Fix issue of not showing directions when switching between origin and destination 

# How
- Set the error messages to not be shown by default when calculating the directions, which was causing the `switch` button to not show any directions because the state was kept as having an error message, and has never been set to `false`.


ammapspeople (approved):
✅

PR #128, opened 2023-04-25
Make the directionsFrom and directionsTo props support using the device position

ammapspeople:
# What

- Make the `directionsFrom` and `directionsTo` props support using the device position.

# How

- Expose the MapsIndoors PositionControl instance up through the components.
- Create hook for holding state for a location for Wayfinding.
  - This location can hold either a MapsIndoors location (given a MapsIndoors Location ID) or the position of the user's device, if setting it as the magic string "USER_POSITION".
  - In case of setting the value to "USER_POSITION", the hook will handle interaction with out MapsIndoors PositionControl instance, and will generate a GoeJSON Feauture with coordinates corresponding the the position.
- Use the hook for setting the internal states for `directionsFromLocation` and `directionsToLocation`.
- Made the Wayfinding component handle the different scenarios.

# Notes for reviewer

- The target branch is the `feature/props-for-directions-from-and-to`


andreeaceachir142 (approved):
✅

andreeaceachir142:
The props `directionsTo` and `directionsFrom` can be used independent of each other, or they always have to be set together?

andreeaceachir142:
Just to clarify, when having both props set to `USER_POSITION` this is the state that the users are in?
image

ammapspeople:
@andreeaceachir142 Yes, that was the agreement with UX. See the discussion in the related ticket.

PR #130, opened 2023-04-26
Add support for tileStyle property

andreeaceachir142:
# What 
- Add support for `tileStyle` property

# How 
- Create a function `onBuildingChanged` and get the default tile url
- Replace the `default` tile style with the tile style name that is provided
- Set the tile URL whenever the floor changes
- Call the function on the `building_changed ` event listener once on the `miInstance`


packages/map-template/README.md
```diff
@@ -62,4 +62,5 @@ To have any Stencil component changes be reflected in this project, you need to
 |`venue`|`string`|The Venue to load from your MapsIndoors Solution. |
 |`locationId`|`string`|Set a MapsIndoors Location ID to show it on the map and its details in the sheet. |
 |`primaryColor`|`string`|The primary color to use throughout the app. |
-|`logo`|`string`|The logo to show during initial load. |
\ No newline at end of file
+|`logo`|`string`|The logo to show during initial load. |
+|`tileStyle`|`string`|Tile style name to change the interface of the map. |
```
ammapspeople, line 66: @markhougaard Can we get your input to the phrasing here?

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -80,16 +87,32 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
         return mapsIndoorsInstance.fitVenue(venue);
     }
 
+    /**
+     * When the building changes, replace the default tile URL style to the incoming tile style.
+     * Set the MapsIndoors Tile URL to the incoming tile style.
+     */
+    const onBuildingChanged = (miInstance) => {
+        if (miInstance) {
+            const tileURL = miInstance.getTileURL().replace('lundefined/', 'l{floor}/').replace(/\/l(\d+)\//gm, '/l{floor}/');
```
ammapspeople, line 96: Maybe you can try to formulate in a comment why the replacement of "lundefined/" is necessary 😄 Try to experiment with what happens if we omit it to get a better understanding.

packages/map-template/README.md
markhougaard: Maybe this?

```suggestion
|`tileStyle`|`string`|Name of Tile Style to display on the map. |
```

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -117,6 +140,20 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
         });
     }
 
+    /*
+     * React on changes in the tile style prop.
+     */
+    useEffect(() => {
+        if (tileStyle) {
+            _tileStyle = tileStyle;
+            onBuildingChanged(mapsIndoorsInstance);
```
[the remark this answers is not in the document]
andreeaceachir142: Renamed function to `onTileStyleChanged` and adjusted description.

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: Simplified the function and removed the unnecessary replacements.

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -12,6 +12,12 @@ const MAP_TYPES = {
 
 const localStorageKeyForVenue = 'MI-MAP-TEMPLATE-LAST-VENUE';
 
+/**
+ * Private variable used for storing the tile style.
+ * Implemented due to the impossibility to use the React useState hook.
+ */
+let _tileStyle;
```
ammapspeople, line 19: Do we need this at all? Can't we just pass the tileStyle to onTileStyleChanged from the useEffect hook?

andreeaceachir142: No, I don't think we need it anymore. I have removed it and used the `tileStyle` instead.

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -80,16 +81,30 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
     }
 
+    /**
+     * Replace the default tile URL style to the incoming tile style.
+     */
+    const onTileStyleChanged = (miInstance) => {
+        if (miInstance) {
+            const tileURL = miInstance.getTileURL().replace('default', tileStyle);
+
+            // Set Tile URL whenever the floor changes
```
ammapspeople, line 91:
```suggestion
            // Replace the floor placeholder with the actual floor and set the tile URL on the MapView.
```

ammapspeople:
In the packages demo, we also set the tile URL whenever the floor changes (the 'floor_changed' event). Do you think we should do the same here?

andreeaceachir142:
Added support for setting the correct tile style whenever the floor changes. ✅

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -80,16 +87,31 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
         return mapsIndoorsInstance.fitVenue(venue);
     }
 
+    /**
+     * Replace the default tile URL style to the incoming tile style.
+     */
+    const onTileStyleChanged = (miInstance) => {
```
ammapspeople, line 93: Can we "exit early" if there is no tileStyle prop set at all? No need to do unnecessary work.

andreeaceachir142: Would a check if `_tileStyle` exists be enough? We could add a check next to the `miInstance`, and then it would only run if we have tile styles.

ammapspeople: Give it a go 👍

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -117,6 +139,20 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
 
+    /*
+     * React on changes in the tile style prop.
+     */
+    useEffect(() => {
+        if (tileStyle) {
+            _tileStyle = tileStyle;
+            onTileStyleChanged(mapsIndoorsInstance);
+        } else {
+            _tileStyle = 'default';
```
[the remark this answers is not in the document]
andreeaceachir142: Moved the `onTileStyleChanged` out of the if statement. However, the assignment of the `default` tile style is necessary, because i.e. we have the `tileStyle=3d_tiles`, but then if we comment/remove the tile style, we need to display the default tile style again.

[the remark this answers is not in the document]
ammapspeople: Got it.

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -117,6 +139,19 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
 
+    /*
+     * React on changes in the tile style prop.
+     */
+    useEffect(() => {
+        if (tileStyle) {
+            _tileStyle = tileStyle;
+        } else {
+            _tileStyle = 'default';
+        }
```
ammapspeople, line 150:
```suggestion
        _tileStyle = tileStyle || 'default' 
```

ammapspeople (approved):
👍

PR #133, opened 2023-05-01
Bugfix/Venue categories

andreeaceachir142:
# What 
- Fix the issue of categories not being shown when non existent venue is set as a property 

# How 
- Pass the `venue` object to the `onVenueChangedOnMap` function on the `Map` component
- If the venue exists, get the categories for that specific venue, otherwise get the categories for the venue that is set on the prop `venue` on `MapsIndoorsMap` component

### DIFF
--- a/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
+++ b/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
@@ -68,20 +68,14 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
     }, [currentAppView]);
 
     /**
-     * When venue is fitted while initializing the data, set map to be ready.
+     * When venue is fitted while initializing the data,
+     * set map to be ready and get the venue categories.
      */
     function venueChangedOnMap(venue) {
         if (isMapReady === false) {
             setMapReady(true);
         }
-
-        // If a venue is passed from the Map component, use that to get the venue categories,
-        // otherwise use the venue that has been set in the prop to get the categories.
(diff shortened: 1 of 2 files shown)


packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -70,11 +70,18 @@ function MapsIndoorsMap({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId
...
-    function venueChangedOnMap() {
+    function venueChangedOnMap(venue) {
         if (isMapReady === false) {
             setMapReady(true);
         }
-        getVenueCategories(currentVenueName);
+
+        // If a venue is passed from the Map component, use that to get the venue categories,
+        // otherwise use the venue that has been set in the prop to get the categories.
+        if (venue) {
```
ammapspeople, line 80: Are we ever in a situation where this is not passed?

andreeaceachir142: Hmm, the venue is always passed. Would it be more correct to check if a `currentVenueName` is set, and if yes, use that, otherwise always default to the `venue` that is passed?

andreeaceachir142: No, we always pass the `venue`. Should we remove the check at all in this case and always assign the `venue.name`?

andreeaceachir142: ✅

ammapspeople (approved):
✅

PR #134, opened 2023-05-01
Bugfix/Distance unit

andreeaceachir142:
# What 
- Fix distance unit to reflect the localization settings on the device

# How 
- Remove hardcoded unit value of `Metric` from the `mi-distance` component used to render the directions steps


ammapspeople:
This breaks the support of the unit prop on the component, so I can no longer explicitly set a unit:

image

andreeaceachir142:
Added support for setting the unit on the component. ✅

ammapspeople (approved):
✅

PR #135, opened 2023-05-01
Bugfix/List item location icon

andreeaceachir142:
# What 
- Fix icon not rendering all the time on the `list-item-location` component

# How 
- Add a `@Watch` to the `icon` property on the `list-item-location` component that will look out for changes in the component and update it correctly, anytime the component changes

### DIFF
--- a/packages/components/src/components/list-item-location/list-item-location.tsx
+++ b/packages/components/src/components/list-item-location/list-item-location.tsx
@@ -29,6 +29,11 @@ export class ListItemLocation {
      */
     @Prop() icon: string;
 
+    @Watch('icon')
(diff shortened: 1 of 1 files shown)


packages/components/src/components/list-item-location/list-item-location.tsx
```diff
@@ -63,7 +68,7 @@ export class ListItemLocation {
      */
     @Event() listItemDidRender: EventEmitter;
 
-    @State() iconURLToRender; string;
+    @State() iconURLToRender: string;
```
ammapspeople, line 71: 😂

ammapspeople (approved):
✅

PR #136, opened 2023-05-01
Bugfix/Live data icon

andreeaceachir142:
# What 
- Add support for showing live data icons 

# How 
- Add condition if the icon source is present to use the `icon.src` to display the icon - covering the live data icons, otherwise use the `icon`


ammapspeople (approved):
👍

PR #137, opened 2023-05-03
bugfix/tile-style-does-not-work-on-google-maps

ammapspeople:
Fix a bug where tiles on Google Maps were never shown. This was caused by the very first call for onTileStyleChanged, where getTileURL would always return null. It is always called later, so the implemented check will not break anything.


andreeaceachir142 (approved):
👍

PR #138, opened 2023-05-04
Bugfix/external id list

andreeaceachir142:
# What 
- Set bottom sheet height when having a list of external IDs to only show 1,5 list items

# How
- Set the height of the `LocationsList` component to be 200px on mobile devices using the media queries 
- Use the `usePreventSwipe` hook to allow the users to scroll the content on the bottom sheet

### DIFF
--- a/packages/map-template/src/components/LocationsList/LocationsList.jsx
+++ b/packages/map-template/src/components/LocationsList/LocationsList.jsx
@@ -2,6 +2,7 @@ import React from "react";
 import './LocationsList.scss';
 import { ReactComponent as CloseIcon } from '../../assets/close.svg';
 import ListItemLocation from "../WebComponentWrappers/ListItemLocation/ListItemLocation";
+import { usePreventSwipe } from "../../hooks/usePreventSwipe";
 
 /**
  * Show list of locations.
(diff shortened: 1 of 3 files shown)


packages/map-template/src/components/LocationsList/LocationsList.jsx
```diff
@@ -21,14 +25,16 @@ function LocationsList({ onBack, onLocationClick, locations }) {
...
             </div>
-            <div className="locations-list__list">
-                {locations?.map(location =>
-                    <ListItemLocation
-                        key={location.id}
-                        location={location}
-                        locationClicked={e => onLocationClick(e)}
-                    />
-                )}
+            <div className="locations-list__scrollable prevent-scroll" {...scrollableContentSwipePrevent}>
```
ammapspeople, line 28: I wonder if this will make the list unscrollable when it should be scrollable: when the sheet is in 100% height and there are more locations than can be shown.

andreeaceachir142: As far as I tested it, when the sheet is in 100% height (max snap point), the list is scrollable even if we have a long list.

ammapspeople: 👏

PR #139, opened 2023-05-04
Add support for changing the props on the MapsIndoorsMap component through URL parameters

andreeaceachir142:
# What 
- Add support for changing the props on the `MapsIndoorsMap` component through the URL parameters

# How 
- Get a hold of the query string and get the values for each property
- Pass them down to the `MapsIndoorsMap` component and add default values where necessary


ammapspeople (approved):
👏

PR #140, opened 2023-05-04
Migrate map template from Create React App to Vite

ammapspeople:
# What

- Migrate map-template from using Create React App to use Vite.

# Why

- In regards to building both React components and web components and publishing to NPM, Vite is a better fit.

# Still missing

- favicon, manifest file etc.


packages/map-template/vite.config.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default defineConfig(() => {
+    return {
+        build: {
+            outDir: 'build'
+        },
+        plugins: [
+           react(),
+           svgr({
+            svgrOptions: {}
```
andreeaceachir142, line 14: The `svgrOptions` are left empty on purpose?

ammapspeople: It wasn't needed. Removed.

andreeaceachir142 (approved):

PR #142, opened 2023-05-09
Add missing information on the directions page 

andreeaceachir142:
# What 
- Add missing information on the origin and destination fields on the Directions page.

# How 
- Store the display rule for the destination and show the icon on the destination field
- Use the `mi-location-info` component to display the information about the origin and destination
- Align styles with the Figma design 
- Add JSDoc for the Directions component


ammapspeople:
There seems to be some issues when using My Position.

a) Weird dot as part of presentation

image

b) The component crashes if I switch "to" and "from" when using My Position and click "Go".

andreeaceachir142:
a) I have fixed the extra dot as part of the presentation, as it should not be shown if the selected location is `My position`. 
b) There is already a ticket for this bug, therefore it will be solved separately. 
@ammapspeople

andreeaceachir142:
I have found another issue, looking into it now. I think it is related to the crashing of the My Position switch as well.
Screen Recording 2023-05-09 at 13.17.05

packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -139,18 +166,41 @@ function Directions({ isOpen, onBack, directions }) {
...
+                        }
                     </div>
                     <div className="directions__container">
                         <label className="directions__label">
                             To
                         </label>
-                        {directions?.destinationLocation && <div>{directions.destinationLocation.properties.name}</div>}
+                        {directions?.destinationLocation &&
+                            <div className="directions__info">
+                                {destinationDisplayRule && directions.originLocation.name !== 'My Position' &&
```
ammapspeople, line 187: I am not sure I understand why originLocation is considered in this part about the destinationLocation. Can you explain?

andreeaceachir142: Oh, that is a mistake. It should be destination location instead of origin. My bad, I have changed it now. 🫣

packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -43,7 +58,19 @@ function Directions({ isOpen, onBack, directions }) {
 
             // Set the step index to be 0 in order to display the correct instruction on the map.
             directionsRenderer.setStepIndex(0);
+
+            originInfoElement.current.location = directions.originLocation;
```
ammapspeople, line 62: Is the useRef needed? Is it because the `<mi-location-info>` doesn't take an object? 

Is it possible to just pass down the `directions.originLocation` to the `<mi-lcoation-info>` element?

andreeaceachir142: @ammapspeople I tried to use the `directions.originLocation` directly on the `mi-location-info` element, but it doesn't seem to be possible. 🤔 Looks like the useRef is needed

ammapspeople: I see that we do the same in LocationDetails as well. All right 😄

ammapspeople (approved):
👏

PR #146, opened 2023-05-12
Feature/remaining query params

andreeaceachir142:
# What 
- Add support for the remaining query parameters

# How 
- Implement support for the `externalIDs`, `tileStyle`, `directionsTo` and `directionsFrom`


packages/map-template/src/App.jsx
```diff
@@ -30,6 +36,10 @@ function App() {
                 primaryColor={hexPrimaryColor}
                 logo={logo ? logo : undefined}
                 appUserRoles={appUserRoles}
+ 				directionsFrom={directionsFrom}
```
ammapspeople, line 39:
```suggestion
                directionsFrom={directionsFrom}
```

PR #147, opened 2023-05-16
Fix for wrong Location shown in the search field after having requested directions

ammapspeople:
# What

- Fix a bug where the `currentLocation` would always persist when going back from Directions to the Wayfinding view.

# How

- Avoid mixing `currentLocation` and `directionsToLocation` in the `to` prop on the Wayfinding component.


andreeaceachir142 (approved):
👏

PR #149, opened 2023-05-16
Fx wrong calculation of directions padding on larger screens.

ammapspeople:
# What

- Fix bug where the left padding for directions on the map was wrongly calculated

# How

- The padding should be the width of the sidebar + padding. It had nothing to do width the width of the map container.
- Also renamed functions for more clarity.


andreeaceachir142 (approved):
👏

PR #150, opened 2023-05-17
Show the location on the map when setting the "locationId" prop

andreeaceachir142:
# What 
- Show the location on the map when setting the `locationId` property 
- Center the map and zoom in until level 21
- Handle the cases when the `locationId` belongs to a different venue
- Handle the cases when the `locationId` is on a different floor than the default one 

# How 
- Pass down the `locationId` property to the `Map` component 
- Resolve the promise instead of fitting the map to the venue when having the `locationId` property set
- Handle the floor selector, the centering of the map and the zooming on the location inside the `building_changed` event listener 
- Set the current venue to be the one that the `locationId` belongs to

### DIFF
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -112,12 +112,15 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
         if (locationId && miInstance) {
             mapsindoors.services.LocationsService.getLocation(locationId).then(location => {
                 if (location) {
+                    // Set the floor to the one that the location belongs to.
                     const locationFloor = location.properties.floor;
                     miInstance.setFloor(locationFloor);
 
+                    // Center the map to the location coordinates.
                     const locationGeometry = location.geometry.type === 'Point' ? location.geometry.coordinates : location.properties.anchor.coordinates;
                     miInstance.getMapView().setCenter({ lat: locationGeometry[1], lng: locationGeometry[0] });
 
+                    // Set the zoom level in order to see the location closer on the map.
                     miInstance?.setZoom(21);
                 }
             });
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -131,10 +131,10 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
     }
 
     /**
-    * Replace the default tile URL style to the incoming tile style.
-    *
-    * @param {object} miInstance
-    */
+     * Replace the default tile URL style to the incoming tile style.
+     *
+     * @param {object} miInstance
+     */
     const onTileStyleChanged = (miInstance) => {
         if (miInstance && _tileStyle) {
             let tileURL = miInstance.getTileURL();
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -99,14 +99,20 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
 
     /**
      * Handle the tile style changes and the locationId property.
-     * If the locationId property is present, set the correct floor, center and zoom the map.
      *
      * @param {object} miInstance
      */
     const onBuildingChanged = (miInstance) => {
-
         onTileStyleChanged(miInstance);
+        onLocationIdChanged(miInstance)
+    }
 
+    /**
+     * If the locationId property is present, set the correct floor, center and zoom the map.
+     *
+     * @param {object} miInstance
+     */
+    const onLocationIdChanged = (miInstance) => {
         if (locationId && miInstance) {
             mapsindoors.services.LocationsService.getLocation(locationId).then(location => {
                 if (location) {
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -61,7 +61,7 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
         if (mapsIndoorsInstance) {
             window.localStorage.clear(localStorageKeyForVenue);
             const venueToShow = getVenueToShow(venueName, venues);
-            if (venueToShow) {
+            if (venueToShow && !locationId) {
                 setVenue(venueToShow, mapsIndoorsInstance).then(() => {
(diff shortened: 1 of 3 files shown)


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -84,18 +85,53 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
+                    // Set the floor to the one that the location belongs to.
+                    const locationFloor = location.properties.floor;
+                    miInstance.setFloor(locationFloor);
+
+                    // Center the map to the location coordinates.
+                    const locationGeometry = location.geometry.type === 'Point' ? location.geometry.coordinates : location.properties.anchor.coordinates;
+                    miInstance.getMapView().setCenter({ lat: locationGeometry[1], lng: locationGeometry[0] });
+
+                    // Set the zoom level in order to see the location closer on the map.
+                    miInstance?.setZoom(21);
```
markhougaard, line 124: @andreeaceachir142 What if a Solution has zoom 22 enabled? Could we fish that out of the Solution data somehow so we zoom is as far as possible?

ammapspeople: Another question is what should happen if the location is a very large room. I imagine in that case we don't want to zoom in as far as possible...?

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -84,18 +85,53 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
      * Set the venue to show on the map.
+     * If locationId property is present, resolve the promise without fitting the venue.
      *
      * @param {object} venue
      * @param {object} mapsIndoorsInstance
      */
     const setVenue = (venue, mapsIndoorsInstance) => {
         window.localStorage.setItem(localStorageKeyForVenue, venue.name);
-        return mapsIndoorsInstance.fitVenue(venue);
+        if (locationId) {
```
ammapspeople, line 95: Would it make sense to make this check "outside" this function? So leave setVenue as it is, but determine if it should be called or not when needed.

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -84,18 +85,53 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
+     * Handle the tile style changes and the locationId property.
+     * If the locationId property is present, set the correct floor, center and zoom the map.
+     *
+     * @param {object} miInstance
      */
+    const onBuildingChanged = (miInstance) => {
+
+        onTileStyleChanged(miInstance);
+
+        if (locationId && miInstance) {
```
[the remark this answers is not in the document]
andreeaceachir142: Refactored the code ✅

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: Yeah, I think so. I refactored the code and checked for the existence of the `locationId` prop whenever we use the `setVenue` function.

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: As agreed within the team, when setting the locationId property, the map will center and zoom in to level 21 for now. 
We have created a ticket for implementing a fitLocation() method on the SDK, and we will implement that in a future iteration, once that is done.

packages/map-template/src/components/Map/Map.jsx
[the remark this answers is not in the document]
ammapspeople: 👀 can't see any changes here

packages/map-template/src/components/Map/Map.jsx
ammapspeople: Does this agreement align with the expectations, @markhougaard ?We will postpone fitting to the geometry and instead just zoom in as much as possible.

packages/map-template/src/components/Map/Map.jsx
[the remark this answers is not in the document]
andreeaceachir142: Latest changes with the refactor weren't pushed 🤷🏻‍♀️ They should be changed now

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -84,6 +87,7 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
 
     /**
      * Set the venue to show on the map.
+     * If locationId property is present, resolve the promise without fitting the venue.
```
ammapspeople, line 90: This is not true here anymore.

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: Yes, there is a task for improving it later on. 🙌🏻

ammapspeople: Then we still need to handle if we should zoom to level 21 or 22. 

Look at the response in the call to `await mapsindoors.services.SolutionsService.getSolution()`. If the `z22` module exists, then we should zoom to level 22.

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: ✅

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: Added support for checking if the solution has `z22` and if yes, set the zoom level to 22, otherwise set it to 21.

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -93,8 +96,52 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
...
+
+                    // Check if the solution allows the zoom level to be 22.
+                    // If yes, set the zoom level to 22, otherwise set it to 21.
+                    mapsindoors.services.SolutionsService.getSolution().then(solution => {
+                        const hasZoom22 = Object.values(solution.modules).find(zoomLevel => zoomLevel === 'z22')
+                        if (hasZoom22) {
+                            miInstance?.setZoom(22);
+                        } else {
+                            miInstance?.setZoom(21);
+                        }
```
ammapspeople, line 134:
```suggestion
                        miInstance?.setZoom(hasZoom22 ? 22 : 21);
```

Will that work?

andreeaceachir142: Yes, it works 👏

PR #151, opened 2023-05-19
Add support for having "startZoomLevel" property

andreeaceachir142:
# What 
- Add support for having a `startZoomLevel` property 

# How 
- Pass the new property `startZoomLevel` to the `Map` component and make the map react to the changes accordingly

### DIFF
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -33,10 +33,11 @@ let _tileStyle;
  * @param {function} props.onMapTypeChanged - Function that is run when the map type is changed.
  * @param {array} props.filteredLocationsByExternalIDs - Array of IDs of the filtered locations based on external ID.
  * @param {string} props.tileStyle - Tile style name to change the interface of the map.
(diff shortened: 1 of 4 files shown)


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -63,10 +64,14 @@ function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocatio
             if (venueToShow) {
                 setVenue(venueToShow, mapsIndoorsInstance).then(() => {
                     onVenueChangedOnMap(venueToShow);
+                    // Set the map zoom level if the property is provided.
```
ammapspeople, line 67: We are calling setVenue twice in this file. Could the logic be moved into the setVenue function?

andreeaceachir142: Moving the logic into the `setVenue` function doesn't work, the `setZoom` seems to not change anything on the map.

ammapspeople: Did you wait for `fitVenue` to resolve before calling `setZoom`?

ammapspeople (approved):

PR #156, opened 2023-05-23
Fix location id property not working on desktop devices

andreeaceachir142:
# What 
- Fix `locationId` property not working on desktop devices.

# How 
- Add another boolean to check if the device is a mobile, and render the `Bottom sheet` and the `Sidebar` according to the correct device.

### DIFF
--- a/packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
+++ b/packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
@@ -33,7 +33,11 @@ function GoogleMapsMap({ gmApiKey, onMapView, onPositionControl, mapsIndoorsInst
             // Initialize Google Maps MapView
             const mapViewOptions = {
                 element: document.getElementById('map'),
-                disableDefaultUI: true // Disable Map Type control, Street view control and Zoom controls.
+                disableDefaultUI: true, // Disable Map Type control, Street view control and Zoom controls.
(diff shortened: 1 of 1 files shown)


ammapspeople (approved):
🎉

packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -33,7 +33,11 @@ function GoogleMapsMap({ gmApiKey, onMapView, onPositionControl, mapsIndoorsInst
             // Initialize Google Maps MapView
             const mapViewOptions = {
                 element: document.getElementById('map'),
-                disableDefaultUI: true // Disable Map Type control, Street view control and Zoom controls.
+                disableDefaultUI: true, // Disable Map Type control, Street view control and Zoom controls.
+                // Always set a center so the map so the bounds or center can be read from the start.
+                center: { lat: 0, lng: 0 },
+                // Set a large zoom so we prevent a "zoom 0 glitch" (showing the whole globe temporarily)
+                zoom: 21
```
andreeaceachir142, line 40: How does that work with the startZoomLevel query param on Google Maps?

ammapspeople: That is set at a later time and should not clash. Just like fitting to the venue will implicitly change the zoom level.

PR #186, opened 2023-05-25
Components: Use shell script when triggering release

markhougaard:
Files are based on a test repo here:

- https://github.com/markhougaard/github-workflow-test/blob/main/packages/components/package.json
- https://github.com/markhougaard/github-workflow-test/blob/main/packages/components/release.sh
- https://github.com/markhougaard/github-workflow-test/blob/main/.github/workflows/components-bump-version.yml

See tags in the test repo here:

- This run: https://github.com/markhougaard/github-workflow-test/actions/runs/5084733450/jobs/9137391524 created this tag: https://github.com/markhougaard/github-workflow-test/releases/tag/components-v33.1.0.

Tested it with the other workspace/package as well: https://github.com/markhougaard/github-workflow-test/actions/runs/5084940648/jobs/9137862560#step:6:29

All of this is because of this bug in the npm-cli: https://github.com/npm/cli/issues/2010 😔

### DIFF
--- a/packages/components/release.sh
+++ b/packages/components/release.sh
@@ -10,7 +10,7 @@ fi
 npm version $1
 git add .
 export NPM_VERSION=$(npm pkg get version --workspaces=false | tr -d \")
-git commit -m 'Release Components'
+git commit -m "Release Components v$NPM_VERSION"
 git tag components-v$NPM_VERSION
 echo "Bumped version to v$NPM_VERSION"
 git push && git push --tags
--- a/packages/components/release.sh
+++ b/packages/components/release.sh
@@ -10,7 +10,7 @@ fi
 npm version $1
 git add .
 export NPM_VERSION=$(npm pkg get version --workspaces=false | tr -d \")
-git commit -m "Release Components v$NPM_VERSION"
-git tag components-v$NPM_VERSION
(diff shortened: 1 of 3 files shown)


packages/components/release.sh
```diff
@@ -0,0 +1,16 @@
...
+    echo "Could not release!"
+    echo "Usage: 'npm run release -- (major|minor|patch)'"
+    echo ""
+    exit 1
+fi
+
+npm version $1
+git add .
+export NPM_VERSION=$(npm pkg get version --workspaces=false | tr -d \")
+git commit -m 'Release Components'
```
ammapspeople, line 13: Did you try something like this to get the version name into the commit message?
https://stackoverflow.com/a/61708683

markhougaard: @ammapspeople Great find; added it now: https://github.com/MapsPeople/web-ui/pull/186/commits/3d42b79afa0942dccbc7e7ffb95f324cbc42e2e4

markhougaard:
@ammapspeople Should we follow a tag structure like the one they have on Astro here: https://github.com/withastro/astro/releases/tag/%40astrojs%2Fmdx%400.19.3 (it's `@astrojs/mdx@0.19.3`, i.e. `@scope/package-name@version`)?

ammapspeople (approved):
✅

PR #194, opened 2023-05-31
Update build workflows

markhougaard:
@andreeaceachir142 @ammapspeople I missed a spot, so it actually didn't run the right `release` command when we merged in earlier today. It means that we didn't tag the commit correctly, but I can do that retroactively. Also, as a side-effect, now we have the world's largest `changelog.md` file, so something good came out of it.

If we label this PR, we'll release new versions of both Components and Map Template, and I'm unsure if we want that because of some internal build process updates.

### DIFF
--- a/.github/workflows/components-bump-version-release.yml
+++ b/.github/workflows/components-bump-version-release.yml
@@ -32,7 +32,7 @@ jobs:
           scope: "@mapsindoors"
       # Build the project to ensure it's up to date. If this steps fails, the job fails.
       - name: Build
-        run: cd packages/components && npm ci && npm run build
+        run: npm ci && npx lerna run build && cd packages/components && npm ci && npm run build
       # You need to set the git config or you can't commit the changes later.
       # The name and email here is for the GitHub Actions user.
       - name: Set git config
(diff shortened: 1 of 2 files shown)


markhougaard:
I tagged the v1.0.0 commit manually now: https://github.com/MapsPeople/web-ui/releases/tag/%40mapsindoors%2Fmap-template%401.0.0

ammapspeople:
@markhougaard If possible, we should avoid releasing new versions to NPM. It seems silly to release new versions without any changes. Who would do that 😼

.github/workflows/components-bump-version-release.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           scope: "@mapsindoors"
       # Build the project to ensure it's up to date. If this steps fails, the job fails.
       - name: Build
-        run: cd packages/components && npm ci && npm run build
+        run: npm ci && npx lerna run build && cd packages/components && npm run build
```
ammapspeople, line 35: Is everything necessary? `npx lerna run build` will run the `npm run build` of all packages.

markhougaard: @ammapspeople What would you suggest it should be then? This? ⤵️

```suggestion
        run: npm ci && npx lerna run build
```

ammapspeople: I like that better, since that seems to make sure than any changes made to MIDT will also be included when building the components.

ammapspeople (approved):
LGTM

andreeaceachir142 (approved):

andreeaceachir142 (approved):

PR #196, opened 2023-06-01
Resize the search view when no categories are shown

andreeaceachir142:
# What 
- Resize the search view when no categories are shown 

# How 
- Add a check if the preferredSizeSnapPoint is undefined and do not set the height of the sheet if that is the case
- Conditionally render the categories if they exist 
- Adjust the styles and the heights of the elements in order to fit the view they need to be rendered in

### DIFF
--- a/packages/map-template/src/components/Search/Search.jsx
+++ b/packages/map-template/src/components/Search/Search.jsx
@@ -150,11 +150,7 @@ function Search({ onLocationClick, categories, onLocationsFiltered, onSetSize, c
      * React on changes in categories prop.
      */
     useEffect(() => {
-        if (categories.length > 0) {
-            setHasCategories(true);
-        } else {
-            setHasCategories(false);
-        }
+        setHasCategories(categories.length > 0);
     }, [categories]);
 
     return (
--- a/packages/map-template/src/components/Search/Search.jsx
+++ b/packages/map-template/src/components/Search/Search.jsx
@@ -29,8 +29,6 @@ function Search({ onLocationClick, categories, onLocationsFiltered, onSetSize, c
     const [searchDisabled, setSearchDisabled] = useState(true);
     const [searchResults, setSearchResults] = useState([]);
 
-    const [hasCategories, setHasCategories] = useState(false);
-
     /** Indicate if search results have been found */
     const [showNotFoundMessage, setShowNotFoundMessage] = useState(false);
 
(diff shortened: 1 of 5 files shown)


packages/map-template/src/components/Search/Search.jsx
```diff
@@ -144,8 +146,21 @@ function Search({ onLocationClick, categories, onLocationsFiltered, onSetSize, c
...
 
+    /*
+     * React on changes in categories prop.
+     */
+    useEffect(() => {
+        if (categories.length > 0) {
+            setHasCategories(true);
+        } else {
+            setHasCategories(false);
+        }
```
ammapspeople, line 157: Would this work as well?

```suggestion
        setHasCategories(categories.length > 0);
```

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -157,30 +172,32 @@ function Search({ onLocationClick, categories, onLocationsFiltered, onSetSize, c
...
+                                key={location.id}
+                                location={location}
+                                locationClicked={e => onLocationClick(e)}
+                            />
+                        )}
+                    </div>
+                }
             </div>
-        </div>
+        </div >
```
ammapspeople, line 200:
```suggestion
        </div>
```

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: Yes, that's nice simplification 👏

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -144,8 +146,17 @@ function Search({ onLocationClick, categories, onLocationsFiltered, onSetSize, c
         }
     }, [currentVenueName]);
 
+    /*
+     * React on changes in categories prop.
+     */
+    useEffect(() => {
+        setHasCategories(categories.length > 0);
```
ammapspeople, line 153: Is this state necessary? Could we just use the categories variable instead of hasCategories, or will changing the categories prop not trigger a re-render of the component?

andreeaceachir142: I have tried using the `categories` variable instead and it seems to be working as intended. ✅

ammapspeople (approved):
👍

PR #197, opened 2023-06-01
Use Recoil to handle global state in the Map Template

ammapspeople:
# What

- Use Recoil to handle global state in the Map Template package.

# Why

- We want to reduce the amount of props being passed up and down the component tree.

# How

- Installed Recoil as dependency.
- Replaced MapsindoorsMap component with basically a Recoil wrapper component and moved the old content to the new MapTemplate component. Review by comparing the old MapsIndoorsMap component from main against the new MapTemplate component.
- Replaced all React contexts and A LOT of component props with Recoil atoms, using either useRecoilValue or useRecoilState


packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -29,7 +30,7 @@ function Directions({ isOpen, onBack, directions }) {
     const [totalDistance, setTotalDistance] = useState();
     const [totalTime, setTotalTime] = useState();
 
-    const mapsIndoorsInstance = useContext(MapsIndoorsContext);
+    const mapsIndoorsInstance = useRecoilValue(mapsIndoorsInstanceState);
```
andreeaceachir142, line 33: Just to make sure I understand correctly the distinction between the `useRecoilState` and `useRecoilValue` - do we use the value here because of the usage of `dangerouslyAllowMutability`?

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -18,40 +34,37 @@ let _tileStyle;
...
-function Map({ apiKey, gmApiKey, mapboxAccessToken, venues, venueName, onLocationClick, onMapsIndoorsInstance, onDirectionsService, onVenueChangedOnMap, onPositionControl, onUserPosition, filteredLocationIds, onMapTypeChanged, filteredLocationsByExternalIDs, tileStyle, startZoomLevel, locationId }) {
-    const [mapType, setMapType] = useState();
-    const [mapsIndoorsInstance, setMapsIndoorsInstance] = useState(null);
+function Map({ onLocationClick, onVenueChangedOnMap }) {
+    const apiKey = useRecoilValue(apiKeyState);
+    const gmApiKey = useRecoilValue(gmApiKeyState);
+    const mapboxAccessToken = useRecoilValue(mapboxAccessTokenState);
+    const [mapType, setMapType] = useRecoilState(mapTypeState);
+    const [mapsIndoorsInstance, setMapsIndoorsInstance] = useRecoilState(mapsIndoorsInstanceState);
+    const [, setUserPosition] = useRecoilState(userPositionState);
```
andreeaceachir142, line 47: Is this the pattern that needs to be followed when using the state, leaving the value empty?

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -248,5 +260,5 @@ function getVenueToShow(preferredVenueName, venues) {
     }
 
     // Else take first venue sorted alphabetically
-    return venues.sort(function (a, b) { return (a.name > b.name) ? 1 : ((b.name > a.name) ? -1 : 0); })[0];
+    return [...venues].sort(function (a, b) { return (a.name > b.name) ? 1 : ((b.name > a.name) ? -1 : 0); })[0];
```
andreeaceachir142, line 263: Why was this change necessary here?

packages/map-template/src/components/Directions/Directions.jsx
ammapspeople: Not quite. We use useRecoilValue when we know we should never update the state from within this component.

packages/map-template/src/components/Map/Map.jsx
ammapspeople: It's one pattern, unsure if is is "the" pattern 😄

packages/map-template/src/components/Map/Map.jsx
ammapspeople: `sort` modifies the venues variable, which is now not allowed due to it being held in a Recoil atom.

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: Would it still work if we would have a value like `userPosition` ?

andreeaceachir142 (approved):

andreeaceachir142:
Approved the latest changes about making the travelMode use Recoil ✅ @ammapspeople

PR #199, opened 2023-06-05
Calculate width of location text based on closing button

andreeaceachir142:
# What 
- Fix issue of closing button not being clickable due to long location name 

# How 
- Calculate the width of the locations info and subtract the width of the closing button in order to avoid overlapping content


ammapspeople (approved):
✅

PR #201, opened 2023-06-06
Fix Wayfinding crashing when having "My position" enabled

andreeaceachir142:
# What 
- Fix Wayfinding crashing when having `My position` enabled

# How 
- Add a null check to the `destinationInfoElement`, and only then set it to de `directions.destinationLocation`


packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -59,7 +59,9 @@ function Directions({ isOpen, onBack, directions }) {
             directionsRenderer.setStepIndex(0);
 
             originInfoElement.current.location = directions.originLocation;
-            destinationInfoElement.current.location = directions.destinationLocation;
+            if (destinationInfoElement.current !== null) {
```
ammapspeople, line 62: How can it be null in the first place? 🤔

andreeaceachir142: Not entirely sure how, but logging it out gives a null value 🤔

ammapspeople: Just worried that we're putting out a small fire and not seeing a bigger fire somewhere else. We should be certain why it is null, because that might reveal some bigger issue.

andreeaceachir142: After looking more into it, it turns out that the issue was an extra unnecessary check when showing the destination location. Removed the check and simplified the code.

PR #202, opened 2023-06-06
Add support for "startZoomLevel" to take priority over "locationId" 

andreeaceachir142:
# What 
- Add support for the `startZoomLevel` to take priority over the `locationId` when present 

# How
- Add a check for the existence of the `startZoomLevel` when zooming into the `locationId`
- If `startZoomLevel` is present, then the map should not zoom to the max zoom level on the selected location


ammapspeople (approved):
✅

PR #203, opened 2023-06-07
Add Floor Selector web component

annakubackaAnku:
Incomplete features:
1. The '1px overlay'
2. The animation could be improved


ammapspeople:
A readme file and a floor-selector.json file is missing.
The readme file will be auto-generated when running the project, and the json file should be manually created.

Some of the content of the readme file should also be auto-generated, but we need to write thorough documentation for each prop available on the component. See other components for inspiration.

annakubackaAnku:
@ammapspeople fixed added ✅

packages/components/src/components/floor-selector/readme.md
```diff
@@ -0,0 +1,102 @@
...
+    const floorSelectorElement = document.createElement('mi-floor-selector');
+    floorSelectorElement.mapsindoors = mi;
+    floorSelectorElement.primaryColor = '#FFC0CB';
+
+    document.body.appendChild(floorSelectorElement);
+```
+
+## `mapsindoors` attribute
+
+A `mapsindoors` attribute is available on the `<mi-floor-selector>` element, which is used to generate the floor-selector.
```
ammapspeople, line 83:
```suggestion
A `mapsindoors` attribute is available on the `<mi-floor-selector>` element, which is required to generate the floor-selector.
```

packages/components/src/components/floor-selector/readme.md
```diff
@@ -0,0 +1,102 @@
...
+
+## `primaryColor` attribute
+
+A `primaryColor` attribute is available on the `<mi-floor-selector>` element, which can be used to set the background color of the selected floor inside the floor-selector.
```
ammapspeople, line 87: Does it have a default value if it is omitted?

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,227 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import { Component, Element, Host, JSX, Prop, State, h } from '@stencil/core';
+
+@Component({
+    tag: 'mi-floor-selector',
+    styleUrl: 'floor-selector.scss',
+    shadow: true
```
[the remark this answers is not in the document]
annakubackaAnku: @ammapspeople we agreed to have an open Shadow DOM at the end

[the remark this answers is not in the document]
ammapspeople: In that case, I think `shadow` should be set to `false` 🤔  Try it out to see if you can apply styling to it from outside the component.

https://stenciljs.com/docs/styling#shadow-dom-in-stencil

packages/components/src/components/floor-selector/readme.md
annakubackaAnku: @ammapspeople yes, I modified it a bit and added the default value

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    /**
+     * Adding/removing classes to the floorlist to fade-in/out the top/bottom part of the container..
+     */
+    private onScrollStyle(): void {
+        const floorListElement = this.el.shadowRoot.querySelector('.mi-floor-selector__list') as HTMLElement;
+        const maxScrollTop = floorListElement.scrollHeight - floorListElement.clientHeight;
+
+        if (this.floorSelectorClosed) {
+            floorListElement.classList.remove('mi-floor-selector__list--fade-top');
```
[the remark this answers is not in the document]
annakubackaAnku: @ammapspeople is it better to put the whole logic of the `onScrollStyle()` function inside the `render()` function?

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,227 @@
...
+    @Element() el: HTMLDivElement;
+
+    private currentFloorElement: HTMLElement;
+    private maxListHeight: number = 300; // The floor-selector.scss: $max-list-height variable needs to be synced with this value
+
+    /**
+     * Scrolling the floorList element to the selected floor.
+     */
+    private scrollToSelectedFloor(): void {
+        const floorSelectorElement = this.el.shadowRoot.querySelector('.mi-floor-selector') as HTMLElement;
```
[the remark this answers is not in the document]
annakubackaAnku: @ammapspeople I updated it for the floorSelectorElement and for the floorListElement

packages/components/src/components/floor-selector/readme.md
```diff
@@ -0,0 +1,102 @@
...
+
+<!-- Auto Generated Below -->
+
+
+## Properties
+
+| Property       | Attribute       | Description | Type     | Default     |
+| -------------- | --------------- | ----------- | -------- | ----------- |
+| `mapsindoors`  | `mapsindoors`   |             | `any`    | `undefined` |
+| `primaryColor` | `primary-color` |             | `string` | `undefined` |
```
ammapspeople, line 97: I don't think the default should be `undefined` here. It's autogenerated, so not sure why it is that way.

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,237 @@
...
+                break;
+            }
+            elementsBeforeSelected.push(floorElements[i]);
+        }
+
+        if (this.floorSelectorClosed) {
+            const listScrollTop = this.floorListElement.scrollTop;
+            const listPaddingTop: number = parseInt(window.getComputedStyle(this.floorListElement).getPropertyValue('padding-top'));
+            const seletedElementDistanceFromTop = this.currentFloorElement.offsetTop - this.floorListElement.offsetTop;
+            const distance: number = -(seletedElementDistanceFromTop - listScrollTop - listPaddingTop);
```
ammapspeople, line 74:
```suggestion
            const distance: number = -(selectedElementDistanceFromTop - listScrollTop - listPaddingTop);
```

ammapspeople: Also can we be more specific than the variable name `distance`?

packages/components/src/components/floor-selector/test.html
```diff
@@ -0,0 +1,61 @@
...
+            }
+        });
+
+        const mapBox = mapView.getMap();
+
+        const floorSelectorElement = document.createElement('mi-floor-selector');
+        floorSelectorElement.mapsindoors = mi;
+        floorSelectorElement.primaryColor = '#FFC0CB';
+
+        document.body.appendChild(floorSelectorElement);
```
ammapspeople, line 57: Use the Mapbox `addControl` method to add it on top of the map. Look in our SDK repository to see how.

ammapspeople:
There's a short scrollbar glitch when opening the floor selector. Can we prevent that?

https://github.com/MapsPeople/web-ui/assets/50314165/9dee29b7-eab2-491d-b8f0-0e9d925b7130

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,237 @@
...
+                        onClick={(): void => this.onToggleFloorSelector()}>
+                    </button>
+
+                    <div
+                        class='mi-floor-selector__list'
+                        ref={(element): HTMLDivElement => this.floorListElement = element as HTMLDivElement}
+                        onScroll={(): void => this.onScrollStyle()}>
+                        {this.floors.map((floor) => (
+                            <button
+                                data-floor={floor.index}
```
ammapspeople, line 207: Can Stencil / JSX help us eliminating the need for the data attribute?

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /**
+     * Adding/removing classes to the floorlist to fade-in/out the top/bottom part of the container..
+     */
+    private onScrollStyle(): void {
+        const maxScrollTop = this.floorListElement.scrollHeight - this.floorListElement.clientHeight;
+
+        if (this.floorSelectorClosed) {
+            this.floorListElement.classList.remove('mi-floor-selector__list--fade-top');
```
ammapspeople, line 128: Can we add and remove classes in the JSX instead of using classList? I imagine the JSX could react to some variable telling what class names to render.

packages/components/src/components/floor-selector/floor-selector.tsx
annakubackaAnku: @ammapspeople I am not sure how, we need the data attribute for styling the floors later on and checking which floor is selected. Did you have a specific solution in mind?

ammapspeople: Nothing specific, no. But using data attributes is a sign that it carries some legacy from "vanilla" JS into a framework, which should help us simplifying stuff like that.

For example, the onClick on the button can just call `this.onSelect` with `floor.index` as parameter instead of passing the click event, which is then used to read the data attribute from the button.

ammapspeople: We discussed this in a meeting today, and agreed that some usages of the data attribute can be changed, but not all.

packages/components/src/components/floor-selector/floor-selector.tsx
annakubackaAnku: @ammapspeople this has also been resolved by introducing 2 properties: `fadeFloorSelectorOnTop` and `fadeFloorSelectorOnBottom` to control the adding/removing of classes

packages/components/src/components/floor-selector/floor-selector.tsx
```diff
@@ -0,0 +1,231 @@
...
+                    .sort((a, b): any => (b as any) - (a as any))
+                    .forEach(floor => {
+                        building.floors[floor].index = floor;
+                        this.floors.push(building.floors[floor]);
+                    });
+            }
+        });
+    }
+
+    /**
```
ammapspeople, line 180: Move this a little further down in the file. I think it's good form to have the render method at the button.

ammapspeople (approved):
👏

PR #210, opened 2023-06-12
Show error message on the search page

andreeaceachir142:
# What 
- Show the error message when no results are found on the search page 

# How 
- Move the error message outside of the search results list container, which is conditioned to be shown only when having search results


ammapspeople (approved):
✅

PR #212, opened 2023-06-16
Improve performance of Splash Screen animations

ammapspeople:
# What & why

Improve the render performance of the animations on the Splash Screen. It appeared to be "janky".

# How

The performance is improved by using a combination of will-change, rotate and transform. This allows the browsers to optimize rendering.


andreeaceachir142 (approved):
✅

PR #213, opened 2023-06-21
Add destination step to the directions

andreeaceachir142:
# What 
- Add destination step to the directions
- Add new type on the route instructions step component for indicating the `Destination` step 

# How 
- Adjust the `mi-route-instructions-step` component to support a new type `destination` for rendering the destination step
- Refactor the `Directions` and the `RouteInstructions` components to use Recoil 
- Create a `useEffect` where we alter the `steps` array by adding the destination step
- Create a `useEffect` where we handle the centering of the map and the zooming in when the user reached the destination step 
- Rewrite the resetting of the active step to 0 when closing the `Directions` page by using the atom `activeStepState`


packages/components/src/components/route-instructions-step/route-instructions-step.scss
```diff
@@ -43,7 +56,7 @@
         @include font.size(large);
         @include margin.all(none);
 
-        > span {
+        >span {
```
ammapspeople, line 59:
```suggestion
        > span {
```

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
```diff
@@ -410,6 +421,27 @@ export class RouteInstructionsStep implements ComponentInterface {
...
+    renderDestinationStep(): JSX.Element {
+        return <div class="step" onClick={e => this.stepClickHandler(e)}>
+            <div class="step__info destination">
+                <div class="step__destination">
+                    {this.isInternetExplorer ? null :
+                        <span class="step__action-icon step__action-icon--circled">
+                            <mi-icon icon-name="logo"></mi-icon>
+                        </span>
+                    }
+                    <div part="step-heading" class="step__heading">You have arrived</div>
```
ammapspeople, line 438: This string should also be part of the `DirectionsTranslations` interface and usage.

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
```diff
@@ -18,22 +24,65 @@ function RouteInstructions({ steps, onNextStep, onPreviousStep, originLocation }
...
+                // Get the destination location
+                const destinationLocation = directions?.destinationLocation;
+
+                // Center the map to the location coordinates.
+                const destinationLocationGeometry = destinationLocation?.geometry.type === 'Point' ? destinationLocation?.geometry.coordinates : destinationLocation?.properties.anchor.coordinates;
+                mapsIndoorsInstance.getMapView().setCenter({ lat: destinationLocationGeometry[1], lng: destinationLocationGeometry[0] });
+
+                // Check if the solution allows the zoom level to be 22.
+                // If yes, set the zoom level to 22, otherwise set it to 21.
+                mapsindoors.services.SolutionsService.getSolution().then(solution => {
```
ammapspeople, line 71: This looks like something I have seen in another part of the code. Can it be extracted into a helper function somewhere?

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
```diff
@@ -18,22 +24,65 @@ function RouteInstructions({ steps, onNextStep, onPreviousStep, originLocation }
...
 
+    /**
+     * Get the zoom and the center of the last step and the destination step in the directions.
+     *
+     */
+    useEffect(() => {
+        if (activeStep === totalSteps?.length - 1) {
+            const lastStepZoom = mapsIndoorsInstance.getZoom();
+            const lastStepCenter = mapsIndoorsInstance.getMapView().getCenter();
+            setLastStepZoom(lastStepZoom);
```
ammapspeople, line 58: Will this be wrong if the user has panned and/or zoomed the map?

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
andreeaceachir142: It is part of the `DirectionsTranslations` interface, on file `directions-translations.interface.ts` on like 33 we have the `destination` string

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
andreeaceachir142: Extracted the code into a helper function ✅

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
andreeaceachir142: Hmm, I don't see why it would be wrong, I have tested it and it seems to be working fine.

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
ammapspeople: But `this.translationsData` is not used to write "You have arrived", so it is not translated.

packages/map-template/src/helpers/SetMapZoomLevel.js
```diff
@@ -0,0 +1,12 @@
+const mapsindoors = window.mapsindoors;
+
+/**
+ * Check if the solution allows the zoom level to be 22.
+ * If yes, set the zoom level to 22, otherwise set it to 21.
+ */
+export default function SetMapZoomLevel(mapsIndoorsInstance) {
```
ammapspeople, line 7:
```suggestion
export default function setMapZoomLevel(mapsIndoorsInstance) {
```

It's usually only React components that has a capital first letter.

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
ammapspeople: See here:

https://github.com/MapsPeople/web-ui/assets/50314165/02282361-30f0-4a87-b680-e4d2bc156bc0

Maybe we should calculate the bounds of the route and use that instead.

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
andreeaceachir142: Oh, I see. How could we add the string so we could use it when displaying the `You have arrived step`? Where is the correct placement for it?

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
andreeaceachir142: That's right, good point. So your idea would be that we calculate the bounds of the route instead of having the lastStepZoom and the lastStepCenter?

ammapspeople: Something like that.

But we can also try to see if the SDK's `setStepIndex` method can be used: https://app.mapsindoors.com/mapsindoors/js/sdk/latest/docs/mapsindoors.directions.DirectionsRenderer.html#setStepIndex

Will it refit the bounds when calling with an index that is the current index?

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
ammapspeople: It should be given as part of the `translations` prop, and then you can use it like:

```suggestion
                    <div part="step-heading" class="step__heading">{this.translationsData.arrived}</div>
```

andreeaceachir142: Added the string to the translations.

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
andreeaceachir142: I tried to use the `setStepIndex` but it doesn't seem to work. I might be doing something wrong - looking into it.

andreeaceachir142: How can we calculate the bound of the route however?

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
```diff
@@ -55,6 +55,12 @@ export class RouteInstructionsStep implements ComponentInterface {
      */
     @Prop() fromTransitStop: string;
 
+    /**
+     * The destination location.
```
ammapspeople, line 59: Should we put a few words in what this is used for?

packages/components/src/components/route-instructions/route-instructions.tsx
```diff
@@ -98,6 +98,7 @@ export class RouteInstructions implements ComponentInterface {
         bike: 'Bike',
         transit: 'Transit',
         drive: 'Drive',
+        destination: 'Destination',
```
ammapspeople, line 101: Should this be "You have arrived"?

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
ammapspeople: Hmmm yeah, I don't think we can in an easy manner, since we also need to take into account if the user has rotated and/or tilted the map.

Another idea could be to find a way to know when the map has panned to the last step. I think the mapView instance will emit an `idle` event whenever the map stops moving. Maybe we can use that.

packages/components/src/components/route-instructions/route-instructions.tsx
andreeaceachir142: Yes ✅

packages/components/src/components/route-instructions-step/route-instructions-step.tsx
andreeaceachir142: ✅

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
andreeaceachir142: Added event listener when map instance is idle ✅

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
```diff
@@ -18,22 +23,64 @@ function RouteInstructions({ steps, onNextStep, onPreviousStep, originLocation }
...
     const [previous, setPrevious] = useState();
 
-    const [activeStep, setActiveStep] = useState(0);
+    const [activeStep, setActiveStep] = useRecoilState(activeStepState);
+    const [totalSteps, setTotalSteps] = useState();
 
-    /*
-     * Make sure to reset active step whenever the steps change.
+    const [lastStepZoom, setLastStepZoom] = useState();
+    const [lastStepCenter, setLastStepCenter] = useState();
```
ammapspeople, line 30: The two useState could be merged into one (object) for performance reasons. We always read and set both of them.

andreeaceachir142: ✅

packages/map-template/src/components/RouteInstructions/RouteInstructions.jsx
```diff
@@ -15,25 +20,66 @@ import { ReactComponent as ArrowLeft } from '../../assets/arrow-left.svg';
...
 function RouteInstructions({ steps, onNextStep, onPreviousStep, originLocation }) {
+
     /** Referencing the previous step of each active step */
     const [previous, setPrevious] = useState();
 
-    const [activeStep, setActiveStep] = useState(0);
+    const [activeStep, setActiveStep] = useRecoilState(activeStepState);
+    const [totalSteps, setTotalSteps] = useState();
+
+    const [lastStep, setLastStep] = useState({ zoom: "", center: "" });
```
ammapspeople, line 30:
```suggestion
    const [lastStepMapState, setLastStepMapState] = useState({ zoom: "", center: "" });
```

ammapspeople (approved):
✅

ammapspeople:
@andreeaceachir142 Reapproved ✅

PR #217, opened 2023-06-28
Add "deploy-to-gcloud" script

markhougaard:
To test this, run the following commands in a shell:

```
$ cd package/map-template
$ npm run deploy-to-gcloud
```

This will fail, because no argument is given.

In the same folder:

```
$ npm run deploy-to-gcloud -- BUCKET_NAME
```

If you have the rights to deploy to that bucket, it will succeed. If not, it will fail.


markhougaard:
Another note: This assumes you have your `.env` file set up properly, and will behave according to what the content of that file is.

ammapspeople (approved):
Looks good to me

andreeaceachir142 (approved):

PR #220, opened 2023-07-03
Add parameter to not show External ID on list item location Web Component

annakubackaAnku:
It would be helpful if there was some handle on this component that may show the external ID by default, but has some parameters that can determine to not show it if set to false.

I added a not required `@Prop()` to the `list-item-location` component called `hideExternalID` - I set the default value to be `false`. The component passes this value to the `location-info` component where the `@Prop() `is also not required and also has the default value of `false`. With this property I extended the check for displaying/hiding the info about the location's external id.

I also did some minor clean-ups for the JSDoc.

### DIFF
--- a/packages/components/src/components/list-item-location/list-item-location.tsx
+++ b/packages/components/src/components/list-item-location/list-item-location.tsx
@@ -51,7 +51,7 @@ export class ListItemLocation {
     }
 
     /**
-     * @description The value of the badge
+     * @description The value of the badge.
      * @type {string} For availability, use "true" or "false".
      */
     @Prop() iconBadgeValue: string;
@@ -81,6 +81,8 @@ export class ListItemLocation {
     private iconDisplaySize = parseInt(midtIcon.icon.size.medium.value);
 
     /**
+     * Emits the location to event listeners.
+     *
      * @description Emits the location to event listeners.
      * @param {*} location - Location object.
      * @memberof List
@@ -89,11 +91,17 @@ export class ListItemLocation {
         this.locationClicked.emit(location);
     }
 
+    /**
+     * Called once just after the component is first connected to the DOM.
+     */
     componentWillLoad(): void {
         this.iconURLToRender = this.icon ? this.icon : this.location?.properties.imageURL;
         this.updateBadge();
     }
 
+    /**
+     * Called after every render().
+     */
     componentDidRender(): void {
         if (!this.location) {
             return;
@@ -109,7 +117,7 @@ export class ListItemLocation {
     }
 
     /**
-     * Apply badge to location icon
+     * Apply badge to location icon.
      */
     updateBadge(): void {
         if (this.iconBadge && this.iconBadgeValue && this.iconURLToRender) {
@@ -118,6 +126,8 @@ export class ListItemLocation {
     }
 
     /**
+     * Set image as background image.
+     *
      * @description Set image as background image.
      * @param {HTMLImageElement} image
      */
@@ -171,6 +181,7 @@ export class ListItemLocation {
 
     /**
      * Create and return an Image from URL.
+     *
      * @param {string} url
      * @returns {Image}
      */
@@ -191,6 +202,8 @@ export class ListItemLocation {
     }
 
     /**
+     * Render location list-item.
+     *
      * @description Render location list-item.
      * @returns {JSX.Element}
      */
@@ -210,6 +223,8 @@ export class ListItemLocation {
     }
 
     /**
+     * Get JSX template for icon.
+     *
      * @description Get JSX template for icon.
      * @returns {JSX.Element}
      */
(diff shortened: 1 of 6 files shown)


packages/components/src/components/list-item-location/list-item-location.tsx
```diff
@@ -76,6 +81,8 @@ export class ListItemLocation {
     private iconDisplaySize = parseInt(midtIcon.icon.size.medium.value);
 
     /**
+     * Emits the location to event listeners.
+     *
      * @description Emits the location to event listeners.
```
ammapspeople, line 86: Seems like we are using the `@description` tag throughout this package. Then there is no need for the added like. The result is the same.

Same for other JSDoc blocks.

https://jsdoc.app/tags-description.html : "If there's both a description at the beginning of a JSDoc comment and a description provided with the @description tag, the description specified with the @description will override the description at the beginning of the comment."

packages/components/src/components/list-item-location/readme.md
```diff
@@ -28,13 +29,18 @@ Example usage:
...
     });
 ```
 
 ## `location` attribute
 
 A `location` attribute is available on the `<mi-list-item-location>` element, which should be used to set the MapsIndoors Location to display information for. The attribute is required.
 
+## `hideExternalID` event
+
+If true, the component will not show the location's Extrenal ID. The attribute is not required.
```
ammapspeople, line 42: Write something about the default value.

packages/components/src/components/list-item-location/test.html
```diff
@@ -61,8 +61,8 @@
     <hr />
 
     <script>
-        const apiKey = 'demo'; // <----- set MapsIndoors API key here
-        document.querySelector('input[name="locationId"]').setAttribute('value', '337006664e044df99916054c');
+        const apiKey = '7315b5a1de3a42f0b379d782'; // <----- set MapsIndoors API key here
```
ammapspeople, line 64: Why is this change needed?

packages/components/src/components/list-item-location/list-item-location.tsx
annakubackaAnku: The code was showing me this warning message almost everywhere where the description was added only with the `@description` tag: "Remove the @description tag to leave a plain block description or add additional description text above the @description line."

ammapspeople: But you're not adding additional description text. It's the same 😁

annakubackaAnku: @ammapspeople I reverted the changes 😁

packages/components/src/components/location-info/location-info.tsx
```diff
@@ -12,15 +12,21 @@ export class LocationInfo implements ComponentInterface {
...
      * Get locations info as a string.
+     *
      * @returns {string}
      */
     getInfoString(): string {
         const details = [];
 
         // External Id
-        if (this.location.properties.externalId) {
+        if (this.location.properties.externalId && !this.hideExternalID) {
```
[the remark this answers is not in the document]
annakubackaAnku: @ammapspeople it makes more sense you're right

packages/components/src/components/list-item-location/readme.md
```diff
@@ -75,13 +81,14 @@ A `listItemDidRender` event is emitted from the `<mi-list-item-location>` elemen
...
-| `location`       | `location`         |             | `any`                                      | `undefined` |
-| `unit`           | `unit`             |             | `UnitSystem.Imperial \| UnitSystem.Metric` | `undefined` |
+| Property         | Attribute          | Description                      | Type                                       | Default     |
+| ---------------- | ------------------ | -------------------------------- | ------------------------------------------ | ----------- |
+| `icon`           | `icon`             |                                  | `string`                                   | `undefined` |
+| `iconBadge`      | `icon-badge`       |                                  | `string`                                   | `undefined` |
+| `iconBadgeValue` | `icon-badge-value` |                                  | `string`                                   | `undefined` |
+| `location`       | `location`         |                                  | `any`                                      | `undefined` |
+| `showExternalId` | `show-external-id` | Whether to show the External ID. | `boolean`                                  | `false`     |
```
ammapspeople, line 90: Default is true?

packages/components/src/components/location-info/readme.md
```diff
@@ -39,21 +40,27 @@ const mockLocation = {
...
 
 ## Properties
 
-| Property   | Attribute  | Description | Type  | Default     |
-| ---------- | ---------- | ----------- | ----- | ----------- |
-| `location` | `location` |             | `any` | `undefined` |
+| Property         | Attribute          | Description | Type      | Default     |
+| ---------------- | ------------------ | ----------- | --------- | ----------- |
+| `location`       | `location`         |             | `any`     | `undefined` |
+| `showExternalId` | `show-external-id` |             | `boolean` | `false`     |
```
ammapspeople, line 63: Default is true?

ammapspeople (approved):
👏

PR #222, opened 2023-07-03
Copy correct build path

markhougaard:
Reading up on the docs for `gsutil`'s `cp` function, the change now uploads the contents correctly.


andreeaceachir142 (approved):
Looks good to me

PR #224, opened 2023-07-03
Fix multiple routes shown simultaneously 

andreeaceachir142:
# What 
- Fix state where multiple routes are shown simultaneously. 

# How 
- Set the directions renderer and the route to null inside the useEffect in order to avoid showing multiple routes on the map.


ammapspeople (approved):
👏

PR #225, opened 2023-07-03
Fix visibility of multiple modals 

andreeaceachir142:
# What 
- Fix visibility of multiple modals shown at the same time.

# How 
- Use the css `display` property in order to only show the active modal.


ammapspeople (approved):
✅

PR #226, opened 2023-07-03
Fix category icons not being shown on the first load 

andreeaceachir142:
# What 
- Fix category icons not being shown on the first load

# How 
- Assign private variable to the app config result


ammapspeople (commented):
We usually do this antipattern/hack when we're trying to circumnvent that the custom events from MapsIndoors are not really playing along with React's lifecycle. But this is not tied to those events at all. I'll dig into looking after another solution.

ammapspeople:
Instead of `useState` or a "private" variable, we can use `useRef`.

```
// Declaration
const appConfigRef = useRef();

// When the App Config has been read from the backend
appConfigRef.current = appConfigResult;

// When we want to use it
const appConfigCategory = appConfigRef.current?.menuInfo.mainmenu.find(category => category.categoryKey === key);
```

andreeaceachir142:
That's a much better solution 👏🏻

ammapspeople:
> That's a much better solution 👏🏻

This pattern may even solve our other "private" hacks.

ammapspeople (approved):
✅

PR #227, opened 2023-07-04
My-position Web Component

annakubackaAnku:
TODO: update script when new SDK is merged in **DONE**


packages/components/src/components/my-position/my-position.scss
```diff
@@ -0,0 +1,66 @@
...
+    position: absolute;
+    pointer-events: auto;
+    gap: 16px;
+    right: 0;
+
+    &__position-button,
+        &__compass-button {
+            margin: 0;
+            cursor: pointer;
+            -webkit-appearance: none;
```
ammapspeople, line 19: I don't think you need the browser prefixes anymore.

https://caniuse.com/?search=appearance

packages/components/src/components/my-position/my-position.tsx
```diff
@@ -0,0 +1,402 @@
...
+import midtOpacity from '@mapsindoors/midt/tokens/opacity.json';
+import { getNestedValue } from '../../utils/utils';
+
+enum PositionStateTypes {
+    POSITION_UNKNOWN = 'POSITION_UNKNOWN',
+    POSITION_REQUESTING = 'POSITION_REQUESTING',
+    POSITION_INACCURATE = 'POSITION_INACCURATE',
+    POSITION_KNOWN = 'POSITION_KNOWN',
+    POSITION_CENTERED = 'POSITION_CENTERED',
+    POSITION_TRACKED = 'POSITION_TRACKED',
```
ammapspeople, line 14: What's the difference between known and tracked? 🤔

packages/components/src/components/my-position/my-position.tsx
```diff
@@ -0,0 +1,402 @@
...
+
+enum DeviceType {
+    Phone = 'mobile',
+    Tablet = 'tablet'
+}
+
+@Component({
+    tag: 'mi-my-position',
+    styleUrl: 'my-position.scss',
+    shadow: true,
```
ammapspeople, line 26: Should this be open for styling from the outside just like we did with the Floor Selector?

packages/components/src/components/my-position/my-position.tsx
```diff
@@ -0,0 +1,402 @@
...
+
+    /**
+     * New UAParser instance.
+     */
+    private parser = new UAParser();
+
+    /**
+     * Wether the MyPositionComponent is a Web Componenet or not.
+     */
+    private readonly isWebComponent: boolean = true;
```
ammapspeople, line 52: This seems unused.

packages/components/src/components/my-position/my-position.tsx
```diff
@@ -0,0 +1,402 @@
...
+        }
+    }
+
+    /**
+     * Rotates and tilts the map.
+     *
+     * @param {DeviceOrientationEvent} e
+     */
+    private handleDeviceOrientation(e: DeviceOrientationEvent): void {
+        if (!this.orientation || this.orientation - (360 - e.alpha) > 1 || this.orientation - (360 - e.alpha) < -1) {
```
ammapspeople, line 194: This difficult to read. Add a comment about what this if statement checks for.

packages/components/src/components/my-position/my-position.tsx
```diff
@@ -0,0 +1,402 @@
...
+                break;
+            case PositionStateTypes.POSITION_UNTRACKED:
+                this.setPositionState(PositionStateTypes.POSITION_TRACKED);
+                this.panToCurrentPosition();
+                break;
+        }
+    }
+
+    /**
+     * Called every time the component is connected to the DOM.
```
ammapspeople, line 337: And what do we do in this callback? 😄 Lots of stuff happening, and it would be nice with a short summary.

packages/components/src/components/my-position/test.html
```diff
@@ -0,0 +1,58 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+   
+    <title>mi-my-position test page</title>
+    <script type="module" src="/build/mi-components.esm.js"></script>
+    <script nomodule src="/build/mi-components.js"></script>
+    <script src="https://polite-terms-watch.loca.lt/build/index.js"></script>
+    <!-- <script src="https://app.mapsindoors.com/mapsindoors/js/sdk/4.23.0/mapsindoors-4.23.0.js.gz?apiKey=mapspeople"></script> -->
```
ammapspeople, line 11: Seems like a test setup 😄

packages/components/src/components/my-position/my-position.tsx
annakubackaAnku: @ammapspeople I couldn't really come up with a better name. 

This is tracked: 
image

This is known:
image

packages/components/src/components/my-position/my-position.tsx
annakubackaAnku: @ammapspeople it isn't used here but it is used by the SDK in the BlueDot files: 

Screenshot 2023-07-05 at 09 10 20

ammapspeople: I'm confused. Are you trying to read a private variable from outside of the component? How does that work?
Anyway, the project cannot be built for production since the linter reports an error about this being unused.

packages/components/src/components/my-position/my-position.tsx
annakubackaAnku: @ammapspeople I talked to @markhougaard and we agreed to have it open, yes

packages/components/src/components/my-position/test.html
annakubackaAnku: @ammapspeople I changed this to use the current version of the SDK but this has to be updated before merging

packages/components/src/components/my-position/my-position.tsx
ammapspeople: Is there another way to determine it, so this property can be removed? 
Can't we just check if the queryselector('mi-my-position') returns an element?

packages/components/src/components/my-position/my-position.tsx
ammapspeople: Ok, that makes more sense.

packages/components/src/components/my-position/my-position.tsx
ammapspeople: 👏 !

packages/components/src/components/my-position/my-position.tsx
ammapspeople: 🙌 !

packages/components/src/components/my-position/my-position.tsx
annakubackaAnku: @ammapspeople good idea, now it is removed ✅

ammapspeople: 🙌

packages/components/src/components/my-position/my-position.json
ammapspeople, line 1: This needs some json content.

ammapspeople (approved):
✅

PR #228, opened 2023-07-04
Change error message on the Wayfinding page

andreeaceachir142:
# What 
- Change error message on the wayfinding page.

# How 
- Remove the error state and only show the `No route found` error message in both scenarios.


ammapspeople (approved):
Consider if this is a minor or a patch.

andreeaceachir142:
Hmmm, it's not really a bug but not a feature either.. Probably a patch would be more suitable.

PR #229, opened 2023-07-04
Load MapsIndoors JavaScript SDK if not already available

andreeaceachir142:
# What 
- Load MapsIndoors JavaScript SDK if not already available 

# How 
- Create a function `initializeMapsIndoorsSDK ` that returns a promise
- Create the script with the latest SDK version if the `mapsindoors` is undefined 
- Replace all the global instances with `window.mapsindoors`
- Refactor the `useEffect` to contain all the other instances where `mapsindoors` is used


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -88,6 +86,89 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+            document.body.appendChild(miSdkApiTag);
+            miSdkApiTag.onload = () => resolve();
+        });
+    }
+
+    /**
+     * Wait for the MapsIndoors JS SDK to be initialized, then react on changes on the props.
+     * React on changes in the MapsIndoors API key by fetching the required data.
+     */
+    useEffect(() => {
```
ammapspeople, line 112: Everything inside this will run every time ANY of the props change. That seems unnecessary, and I didn't think this through when we discussed the solution earlier.

Could we instead have a state hook, for example `mapsindoorsSDKAvailable` that is set to true when the onload for the script resolves?

And then you could add that as dependency on the smaller useEffects and only run the original code if the SDK is available.

andreeaceachir142: Refactored the `useEffect` and added `mapsindoorsSDKAvailable` state. Should we have a default value for it?

ammapspeople: I guess the default should be false until we make the check and resolve.

ammapspeople:
Is it possible to revert the reordering of the useEffect calls, so the diff will be easier to read? Or is the reordering intentional?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Added default value to be false

andreeaceachir142:
The reordering was intentional with the purpose of easier readability - before the useEffects were mixed up with other functions. Now I ordered the ones dependent on the `mapsindoorsSDKAvailable` first, and the rest after. 
If it's difficult to read, I can go back and try to revert the order.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -88,6 +86,133 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+            }
+        });
+    }
+
+    /**
+     * Wait for the MapsIndoors JS SDK to be initialized, then react on changes on the props.
+     * React on changes in the MapsIndoors API key by fetching the required data.
+     */
+    useEffect(() => {
+        initializeMapsIndoorsSDK().then(() => {
```
ammapspeople, line 120: Could the call be done in a useEffect with [] as dependency? I feel like that would be more clean: Why should exactly this useEffect for the apiKey be responsible for initializing the SDK?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -88,6 +86,133 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+    /*
+     * React on changes in the externalIDs prop.
+     * Get the locations by external IDs, if present.
+     */
+    useEffect(() => {
+        if (externalIDs && mapsindoorsSDKAvailable) {
+            window.mapsindoors.services.LocationsService.getLocationsByExternalId(externalIDs).then(locations => {
+                setFilteredLocationsByExternalID(locations);
+            });
+        } else {
```
ammapspeople, line 177: Do we really want to call the set function when there is no mapsindoors SDK available?
I think everything could be wrapped in a check for `mapsindoorsSDKAvailable`, just like the useEffect for locationId.

andreeaceachir142: Good idea. I refactored the useEffect for the external IDs.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Hm, just to clarify that I understood correctly. Should we have a separate useEffect for the `initializeMapsIndoorsSDK` and then use the `mapsindoorsSDKAvailable` as a dependency on the useEffect for the apiKey?

ammapspeople: Sorry for being unclear. But yes 😄

andreeaceachir142: I have refactored the `useEffect` - is that the solution that you had in mind?

ammapspeople: Yup 👍

ammapspeople (approved):

andreeaceachir142:
Upgraded the SDK version to 4.24.0 as well @ammapspeople

ammapspeople:
I also think the README file should be updated so the code example no longer requires the SDK script tag. I am sure @markhougaard will be thrilled about requiring one less line of code for the Map Template to run.

ammapspeople:
Also: we can remove the SDK script tag from index.html now.

andreeaceachir142:
Oh, alright. I removed the SDK script from the index.html file

PR #230, opened 2023-07-04
Update npm link for midt

andreeaceachir142:
# What 
- Update npm link for midt

### DIFF
--- a/packages/midt/package.json
+++ b/packages/midt/package.json
@@ -28,5 +28,5 @@
     "np": "^7.6.3",
     "style-dictionary": "^3.7.2"
   },
-  "homepage": "https://github.com/MapsPeople/web-ui"
+  "homepage": "https://github.com/MapsPeople/web-ui/tree/main/packages/midt"
 }
--- a/packages/midt/package.json
+++ b/packages/midt/package.json
@@ -6,7 +6,8 @@
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "git://github.com/MapsPeople/web-ui.git"
+    "url": "https://github.com/MapsPeople/web-ui.git",
+    "directory": "packages/midt"
   },
   "keywords": [
     "mapsindoors",
--- a/packages/midt/package.json
+++ b/packages/midt/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "git://github.com/MapsPeople/web-ui/tree/main/packages/midt.git"
+    "url": "git://github.com/MapsPeople/web-ui.git"
   },
   "keywords": [
     "mapsindoors",
--- a/packages/midt/package.json
+++ b/packages/midt/package.json
@@ -6,7 +6,7 @@
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


andreeaceachir142:
Not sure about updating the homepage link for the midt, I can't see any `homepage` value on the package.json file for midt. Is that something that happens automatically somehow? @markhougaard 
https://www.npmjs.com/package/@mapsindoors/midt

packages/midt/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "git://github.com/mapspeople/midt.git"
+    "url": "git://github.com/MapsPeople/web-ui/tree/main/packages/midt.git"
```
markhougaard, line 9:
```suggestion
    "url": "git://github.com/MapsPeople/web-ui.git"
```

With midt being part of the generic web-ui repo, I'd guess this is the git url to put here, right?

markhougaard (changes requested):

packages/midt/package.json
andreeaceachir142: If we want to point the users to the `web-ui` only instead of the full path to the `midt`, then yes

andreeaceachir142:
Alright, added web-ui as the homepage.

packages/midt/package.json
markhougaard: The `midt` folder is part of the `web-ui` git repo, so there is no such thing as `midt.git`, but we could link to its folder as the homepage.

andreeaceachir142:
Alright, so I updated the homepage to point to the midt folder and the repository to point to web-ui. @markhougaard

packages/midt/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "git://github.com/mapspeople/midt.git"
+    "url": "git://github.com/MapsPeople/web-ui.git"
```
markhougaard, line 9: Alright, I wanted to educate myself a bit, and see if there was a smarter way to do this 😀

So, I checked out the React Native repo for inspiration (I just wanted a popular one), and found this:

[`"repository": "github:facebook/react-native",`](https://github.com/facebook/react-native/blob/0c9c57a9f73294414d92428c5d2472dc1e1e5e96/package.json#L7)

The shorthand for `github:mapspeople/web-ui` would be nice, but there is an even better way.

Reading the docs for `package.json` on npm, they have this example:

```
{
  "repository": {
    "type": "git",
    "url": "https://github.com/facebook/react.git",
    "directory": "packages/react-dom"
  }
}
```

Which is how you link to a package inside a monorepo 🎉 

So, my suggestion is to have these lines like so (but I can't suggest it because multi-line suggestions on deleted lines is not possible here 🤷):

```
  "repository": {
    "type": "git",
    "url": "https://github.com/MapsPeople/web-ui.git",
    "directory": "packages/midt"
  }
```

andreeaceachir142: That's pretty nice! 🥳 I updated the package.json with the suggested changes.

markhougaard (approved):

PR #236, opened 2023-07-11
Replace mapsindoors global variable

andreeaceachir142:
# What 
- Replace mapsindoors global variable

# Why 
- A previous merge which took care of loading the MapsIndoors JS SDK if not available changed the way we use the mapsindoors, and we had to replace the instances where the global variable was used. Some of the instances were not changed and therefore threw errors 

# How 
- Remove all the global variables for the `window.mapsindoors`


annakubackaAnku:
Nice! 🥳 It runs perfectly for me✅

annakubackaAnku:
Should the Changelog be updated?

annakubackaAnku (approved):

PR #238, opened 2023-07-14
Add support for having URL parameters on the web component 

andreeaceachir142:
# What 
- Add support for having URL parameters on the web component 

# How 
- Moved all the logic from the `App.jsx` file which handles the URL parameters
- Create a new prop `supportsUrlParameters` which indicates a voluntary choice of the Map Template supporting the URL Parameters
- Add the logic for supporting the URL parameters in every useEffect for each prop
- Cover the 6 scenarios when having the URL parameters.

- Example for handling the `apiKey` prop: 
1. The map template supports URL parameters and has an `apiKey` URL parameter set 
2. The map template supports URL parameters and does not have the `apiKey` URL parameter set 
3. The map template supports URL parameters and does not have any of the `apiKey` prop or the `apiKey` URL parameter set
4. The map template supports URL parameters and has both the `apiKey` URL parameter and the `apiKey` prop set
5. The map template does not support URL parameters and has the `apiKey` prop set
6. The map template does not support URL parameters and does not have any `apiKey` prop set 
- Created a useEffect that handles the `logo` prop and created an atom for its state, therefore we could get rid of passing the prop down to the Splash screen component 
- I have tried to create a work around for the `directionsTo` and `directionsFrom` in order to use Recoil and atoms, and use them in an useEffect, but could not quite get it to work properly - I have a stash which I can share, just to see my approach 
- Created two functions that return the location for handling the `directionsTo` and `directionsFrom` - the only thing here is that even now they are not handled in a useEffect, so any changes runtime will be disregarded 

**Note**
- The handling of the URL parameters added quite a large amount of code to the `MapTemplate` component, so I am thinking if we could optimise it in any way 
- I have tried to move everything in a single useEffect and handle the URL parameters there, but then I thought that it might not be such a good idea because that will run every single time one of the prop changes


ammapspeople:
Would it make sense to have the URL param handling live in the MapsIndoorsMap component instead of the MapTemplate component?
It's the MapsIndoorsMap component that we export as a web component, so maybe it fits better there.

ammapspeople:
Remember to add the new prop to the MapsIndoorsMap.propTypes object.

PR #239, opened 2023-07-14
Hide floor selector when directions are open

andreeaceachir142:
# What 
- Hide floor selector when directions are open

# How
- Due to the refactoring of the floor selector in the SDK, the class has been changed from `floor-selector` to `mi-floor-selector`, therefore the styling has been lost because the SCSS file has not been updated with the new class


ammapspeople (approved):
✅

PR #240, opened 2023-07-17
Refactor private variable

andreeaceachir142:
# What 
- Refactor private variable 

# How 
- Replace the private variable with the `userRef` and assign the boolean to the reference instead of the private variable


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -230,7 +227,7 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         }
 
         setHasDirectionsOpen(currentAppView === appStates.DIRECTIONS);
-        _locationsDisabled = currentAppView === appStates.DIRECTIONS;
+        locationsDisabledRef.current = currentAppView === appStates.DIRECTIONS;
```
ammapspeople, line 230: Do we need both the `locationsDisabled` useRef and the `hasDirectionsOpen` useState? They contain the same information.

andreeaceachir142: Good point, we might be able to get rid of one of them. In this case, should we keep the useRef instead of the useState?

ammapspeople: I think the useRef works better.

andreeaceachir142: Refactored the code and removed the `hasDirectionsOpen` state.

ammapspeople:
Is this more a patch release than a minor release?

andreeaceachir142:
Actually yeah, I think it's a patch rather than a minor

ammapspeople (approved):
✅

PR #241, opened 2023-07-17
Hide visibility of external ids by default

andreeaceachir142:
# What 
- Hide visibility of external ids by default 

# How 
- Set the boolean which is responsible for showing/hiding the external ids to false


ammapspeople (approved):
👏

PR #242, opened 2023-07-17
Fix scrollable list on Safari

andreeaceachir142:
# What 
- Fix scrollable list not respecting the height on Safari

# How 
- Remove the `min-height` property from the modal, which Safari interprets differently than other browsers
- Tested it on other browsers as well and it works as intended


ammapspeople (approved):
✅

PR #243, opened 2023-07-18
Fix my position button not working in the wayfinding

andreeaceachir142:
# What 
- Fix my position button not working in the wayfinding view

# How 
- Add missing `detail` parameter when creating the position control

### DIFF
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -212,8 +212,8 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
             });
         } else {
             positionControl.on('position_received', positionInfo => {
-                if (positionInfo.detail.accurate === true) {
-                    setUserPosition(positionInfo.detail.position);
+                if (positionInfo.accurate === true) {
+                    setUserPosition(positionInfo.position);
                 }
             });
         }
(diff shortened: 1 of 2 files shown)


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -206,14 +206,14 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
...
+                if (positionInfo.detail.accurate === true) {
+                    setUserPosition(positionInfo.detail.position);
                 }
             });
         } else {
             positionControl.on('position_received', positionInfo => {
-                if (positionInfo.accurate === true) {
-                    setUserPosition(positionInfo.position);
+                if (positionInfo.detail.accurate === true) {
+                    setUserPosition(positionInfo.detail.position);
```
annakubackaAnku, line 216: we don't need the `.detail` here - the `else` part of the main condition is checking if the my-position button is put in as an mi-component or if it is using the SDK's standard my-position button. In the latter case, the `positionInfo.accurate` and  `positionInfo.position` should work.

andreeaceachir142: Oh, alright. Removed the unnecessary part

annakubackaAnku (approved):

annakubackaAnku (approved):

PR #244, opened 2023-07-25
MIAJS-1748 feature/matb: Add 'Level' word translation

matbmapspeople:
I created a property with default value of 'Level'.
When we bind to it inside kiosk we will pipe it through translation for languages that we support with translations.

### DIFF
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -360,6 +360,10 @@ export namespace Components {
     }
     interface MiLocationInfo {
         /**
+          * @description Level info.
+         */
+        "level": string;
+        /**
           * @description Location object.
          */
         "location": any;
@@ -1698,6 +1702,10 @@ declare namespace LocalJSX {
     };
     }
     interface MiLocationInfo {
+        /**
+          * @description Level info.
+         */
+        "level"?: string;
         /**
           * @description Location object.
          */
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -360,7 +360,7 @@ export namespace Components {
     }
     interface MiLocationInfo {
         /**
-          * @description Level info.
+          * @description The word used for "Level" when showing level info. Default is "Level".
          */
         "level": string;
         /**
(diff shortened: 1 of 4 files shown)


packages/components/src/components/location-info/location-info.tsx
```diff
@@ -12,6 +12,11 @@ export class LocationInfo implements ComponentInterface {
      */
     @Prop() location;
 
+    /**
+     * @description Level info.
```
ammapspeople, line 16: "info"? 🤔 Isn't it just a word or a phrasing?

matbmapspeople: @ammapspeople fixed ✅

ammapspeople: I don't think I would understand what it was about if I was new to using the component. Can you elaborate the description to be more explaining than just "Level"?

matbmapspeople: I tried to be more descriptive now!

packages/components/src/components.d.ts
```diff
@@ -1698,6 +1702,10 @@ declare namespace LocalJSX {
     };
     }
     interface MiLocationInfo {
+        /**
+          * @description Level info.
```
ammapspeople, line 1706: Run build scrips (or whatever) to make these autogenerated files reflect that is written in the tsx file.

packages/components/src/components/location-info/location-info.tsx
```diff
@@ -12,6 +12,11 @@ export class LocationInfo implements ComponentInterface {
      */
     @Prop() location;
 
+    /**
+     * @description Location's level property that can by translatable for specific languages. By default level value is 'Level'.
```
ammapspeople, line 16: Make the "would grandma understand this" test 😄 
The phrasing "Location's level property" is confusing.
How about

`The word used for "Level" when showing level info. Default is "Level".`

matbmapspeople:
Suggestions applied.

ammapspeople:
Remember to apply a "minor" label to this pull request. That will make sure it is released to NPM as a semantically minor version (it is an addition after all).

ammapspeople (approved):

PR #245, opened 2023-07-28
Substeps timing issue when going back from last step.

ammapspeople:
refactor: move away from storing map center and zoom level in order to reset bounds when going back from destination step. Instead, trigger the SDK to refit the bounds.


andreeaceachir142 (approved):
✅

andreeaceachir142:
Very nice refactor! 👏🏻

PR #247, opened 2023-08-01
Fix error messages appearing in weird states

andreeaceachir142:
# What 
- Fix error messages appearing in weird states 
- Fix error message appearing when Google places are present 

# How 
- Set the correct states when getting Google places and when displaying the error messages


ammapspeople (approved):
✅

PR #248, opened 2023-08-01
Handle props and URL params in MapsIndoorsMap component

ammapspeople:
# What

Refactor the handling of props and query parameters to be the responsibility of the MapsIndoorsMap component.

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -51,9 +51,8 @@ let _locationsDisabled;
  * @param {array} [props.externalIDs] - Filter locations shown on the map based on the external IDs.
  * @param {string} [props.tileStyle] - Tile style name to change the interface of the map.
  * @param {number} [props.startZoomLevel] - The initial zoom level of the map.
- * @param {boolean} [props.hasURLParameters] - If you want to support URL Parameters.
  */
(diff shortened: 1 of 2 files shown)


andreeaceachir142 (approved):
✅

andreeaceachir142:
There are still some places where we use `hasURLParameters` on the MapTemplate.jsx file. Do we still need that, or can we get rid of it completely on that file?

andreeaceachir142:
Do we need all the props defined on the MapsIndoorsMap.jsx component, or is it enough to have them on the MapTemplate.jsx?

ammapspeople:
> Do we need all the props defined on the MapsIndoorsMap.jsx component, or is it enough to have them on the MapTemplate.jsx?

The MapsIndoorsMap is the "public facing" component (and web component), so it needs to be there and passed down to the child component, MapTemplate. Or am I misunderstanding? Can you elaborate?

andreeaceachir142:
> > Do we need all the props defined on the MapsIndoorsMap.jsx component, or is it enough to have them on the MapTemplate.jsx?
> 
> The MapsIndoorsMap is the "public facing" component (and web component), so it needs to be there and passed down to the child component, MapTemplate. Or am I misunderstanding? Can you elaborate?

The same props are defined on the MapTemplate and also on the MapsIndoorsMap files - and I thought they don't need to be present in both files, but it makes sense that they need to exist on the public facing component and web component too. 

image

ammapspeople:
> There are still some places where we use `hasURLParameters` on the MapTemplate.jsx file. Do we still need that, or can we get rid of it completely on that file?

It was my ambition to only have that in the MapsIndoorsMap component. I'll look into it.

ammapspeople:
@andreeaceachir142 I removed the unnecessary hasURLParameters from the MapTemplate component now.

PR #249, opened 2023-08-02
Highlight with mouse should not close dropdown

matbmapspeople:
The scope of this task is to prevent dropdown to close when highlighting content inside input field, while mouse click is released outside the input field.

**How does it look like now:**

https://github.com/MapsPeople/web-ui/assets/56337591/08c51df1-b3bb-448a-b47a-29ee4a63887e

My implementation proposition:

Inside `checkForClickOutside()`:
- I created couple of consts: for listening clicks inside hostElement or inside canvas (map), also const that checks if anything inside an input is highlighted with mouse.
- Then I made if checks.

I tried to be descriptive with JSDocs, but let me know if you have any questions.

**My implementation in action:**

https://github.com/MapsPeople/web-ui/assets/56337591/85e96e8a-5772-49ef-8bfd-eab12d3d39b0

### DIFF
--- a/packages/components/src/components/combo-box/combo-box.tsx
+++ b/packages/components/src/components/combo-box/combo-box.tsx
@@ -137,7 +137,7 @@ export class ComboBox {
      * @param {ev} ev
      */
     @Listen('mousedown', { target: 'window' })
-    clearSelectionOnClick(ev) {
+    clearSelectionOnMouseDown(ev) {
         if (!this.hostElement.contains(ev.target) && ev.target !== this.svgButtonToggleDropdown) {
             window.getSelection().empty();
         }
--- a/packages/components/src/components/combo-box/combo-box.tsx
+++ b/packages/components/src/components/combo-box/combo-box.tsx
@@ -496,7 +496,6 @@ export class ComboBox {
         return (
             <Host class={{ 'open': this.open }}>
                 <input
-                    id='comboBoxInputField'
                     type="text"
                     class="input"
                     onFocus={(event) => this.onClickExists(event)}
--- a/packages/components/src/components/combo-box/combo-box.tsx
+++ b/packages/components/src/components/combo-box/combo-box.tsx
@@ -16,6 +16,7 @@ enum KeyCode {
 })
 export class ComboBox {
     @Element() hostElement: HTMLElement;
+    // @Element() canvasElement: HTMLCanvasElement;
 
     /**
      * Triggers an event when the selection is changed.
@@ -122,6 +123,7 @@ export class ComboBox {
     isMouseOverEventDisabled = false;
 
     filterElement!: HTMLInputElement;
+    svgButtonToggleDropdown!: SVGElement;
     clearButtonElement!: HTMLButtonElement;
     listElement!: HTMLElement;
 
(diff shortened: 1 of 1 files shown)


packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -122,24 +123,72 @@ export class ComboBox {
...
+    svgButtonToggleDropdown!: SVGElement;
     clearButtonElement!: HTMLButtonElement;
     listElement!: HTMLElement;
 
     currentItemIndex = 0;
     selectedItemIndex = 0;
     highlightedItemClassName = 'list__item--highlighted';
 
+    /**
+     * If we click on the map (canvas) we want to empty selection, so this.open can be set to false.
```
ammapspeople, line 135: What map are we referring to here?

packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -122,24 +123,72 @@ export class ComboBox {
...
      */
     @Listen('click', { target: 'window' })
+    @Listen('mouseup', { target: 'window' })
     checkForClickOutside(ev) {
-        if (!this.hostElement.contains(ev.target) && this.noResultsMessage) {
-            this.clearFilter();
-            this.filterElement.value = this.selected[0]?.text;
-            this.open = false;
+        // Check if the event target is inside the input field or canvas
+        const clicksInsideBody = document.getElementsByTagName('body');
```
ammapspeople, line 155: I don't think I understand the naming of this variable. `getElementsByTagName('body')` will return a collection of HTML elements (and how can multiple body elements exist in the document? Can't `document.body` be used instead).

How does that relate to something with clicks?

packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -122,24 +123,72 @@ export class ComboBox {
...
+        const isTargetInsideInput = this.hostElement.contains(ev.target);
+        const isTargetInsideCanvas = clicksInsideBody[0].contains(ev.target);
```
ammapspeople, line 157: What is "canvas" here? `clicksInsideBody[0]` refers to a `body` element.

packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -122,24 +123,72 @@ export class ComboBox {
...
+
+        // Check if there is any selected text inside the input field
+        const selection = window.getSelection().toString();
+        const hasSelectedText = selection.length > 0;
```
ammapspeople, line 161:
```suggestion
        const hasSelectedText = window.getSelection().toString().length > 0;
```

packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -122,24 +123,72 @@ export class ComboBox {
...
+
+        const inputInFocus = this.isComboboxInputFieldInFocus('menu__combobox');
+
+        // If we click outside an input field but the text was highlihted with the mouse, dropdown will remain opened.
```
ammapspeople, line 165:
```suggestion
        // If we click outside an input field but the text was highlighted with the mouse, dropdown will remain opened.
```

packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -122,24 +123,72 @@ export class ComboBox {
...
+     * @returns {boolean}
+     */
+    isComboboxInputFieldInFocus(inputFieldClassName: string): boolean {
+        const focusedElement = document.activeElement;
+
+        if (focusedElement.className === inputFieldClassName) {
+            return true;
+        } else {
+            return false;
         }
```
ammapspeople, line 192:
```suggestion
        return focusedElement.className === inputFieldClassName;
```

matbmapspeople:
Closed due to no more need to this complicated fix.

PR #250, opened 2023-08-03
Upgrade to MapsIndoors SDK v4.24.4

ammapspeople:
Upgrade MapsIndoors SDK to newest version and remove behaviour that is now default in the SDK.


andreeaceachir142 (approved):
✅

PR #251, opened 2023-08-03
Fix startZoomLevel property not working on Google Maps

andreeaceachir142:
# What 
- Fix `startZoomLevel` property not working on Google Maps

# Why 
- The use of `startZoomLevel` on a Google Maps map was not working and it threw the error: `InvalidValueError: setZoom: not a number`

# How 
- After some investigation on the error that was being thrown, I converted the `startZoomLevel` value into an integer when setting the zoom of the map using the `parseInt()` method.


ammapspeople (approved):
Nice ✅

PR #253, opened 2023-08-07
Make a "use my position" option on the Wayfinding

andreeaceachir142:
# What 
- Make a "use my position" option on the Wayfinding 

# How 
- Create a function which indicates which search field has been focused, and show the `My position` option when the search is performed
- Create a new state which indicates if any of the search fields already have the `My position` option selected, because the users can ONLY select the `My position` option in one of the search fields


packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -404,16 +435,18 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
                 </div>
             </div>
             {!hasFoundRoute && <p className="wayfinding__error">No route found</p>}
             {!hasSearchResults && <p className="wayfinding__error">Nothing was found</p>}
             {searchResults.length > 0 &&
                 <div className="wayfinding__scrollable" {...scrollableContentSwipePrevent}>
                     <div className="wayfinding__results">
+                        {userPosition && !hasMyPostion && <div className="wayfinding__use-current-position" onClick={() => selectMyPosition()}>
+                            <CompassArrow />
+                            <span>My Position</span>
```
ammapspeople, line 448: Is the span necessary?

```suggestion
                            My Position
```

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -172,17 +176,26 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
     }
 
     /**
-     * Set the user's current position as the origin.
-     *
-     * This is done by having a GeoJSON Feature with geometry corresponding to
-     * the user's position.
-     */
```
ammapspeople, line 179: Revert indentation change.

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -92,6 +93,9 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
 
     const [travelMode, setTravelMode] = useRecoilState(travelModeState);
 
+     /** Indicate if the user has My Position selected */
+    const [hasMyPostion, setHasMyPosition] = useState(false);
```
ammapspeople, line 97: `has` can also indicate that the position is known by the app. How about `myPositionSelected`?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: ✅

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: Good point, renamed the use state ✅

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -404,16 +440,19 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
                     </label>
-                    {userPosition && originLocation?.id !== 'USER_POSITION' && <p className="wayfinding__use-current-position">
-                        <button onClick={() => setMyPositionAsOrigin()}>Use My Position</button>
-                    </p>}
+
                 </div>
             </div>
             {!hasFoundRoute && <p className="wayfinding__error">No route found</p>}
             {!hasSearchResults && <p className="wayfinding__error">Nothing was found</p>}
+            {userPosition && !myPositionSelected && showMyPosition && <div className="wayfinding__use-current-position" onClick={() => selectMyPosition()}>
```
ammapspeople, line 448: Shouldn't this be a `<button>` instead?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -92,6 +93,12 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
 
     const [travelMode, setTravelMode] = useRecoilState(travelModeState);
 
+    /** Indicate if the user has My Position selected */
+    const [myPositionSelected, setMyPositionSelected] = useState(false);
+
+     /** Indicate if the option to choose My Position should be shown */
+    const [showMyPosition, setShowMyPosition] = useState(true);
```
ammapspeople, line 100: This can also indicate that the position is shown. 
It's just a button to tell the directions to use that position, right?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: Yeah, replaced it to be a button.

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -404,16 +440,19 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
                     </label>
-                    {userPosition && originLocation?.id !== 'USER_POSITION' && <p className="wayfinding__use-current-position">
-                        <button onClick={() => setMyPositionAsOrigin()}>Use My Position</button>
-                    </p>}
+
                 </div>
             </div>
             {!hasFoundRoute && <p className="wayfinding__error">No route found</p>}
             {!hasSearchResults && <p className="wayfinding__error">Nothing was found</p>}
+            {userPosition && !myPositionSelected && showMyPosition && <button className="wayfinding__use-current-position" onClick={() => selectMyPosition()}>
```
ammapspeople, line 448:
```suggestion
            {userPosition && !myPositionSelected && showMyPosition && <button type="button" className="wayfinding__use-current-position" onClick={() => selectMyPosition()}>
```

just in case the button is suddenly put inside a form element 😄

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: We need a way to distinguish between the state when the My Position option has been selected and when the My Position option should be shown. I tried to merge those to into one single state, but could not get it to work. 

We use the `showMyPosition` because we would like to show the option even when a search has not been yet performed. So therefore we use it when the search field has been clicked, the my position option should be shown, and we cannot set the `myPositionSelected` state in that case. 

The same applies when a location has been clicked, we need to use the `showMyPosition` in that case, otherwise we get into a state where we have both the My Position option and the details for the selected location.

And we should also use the `showMyPosition` when there are no results. 

Let me know if it doesn't make sense or if you have any other suggestion of how to approach this. 😄

ammapspeople: Naming is hard! Could it be `showMyPositionOption`?

andreeaceachir142: Yeah, good idea actually! I changed the name.

ammapspeople (approved):
✅

packages/components/src/components/search/search.tsx
```diff
@@ -233,6 +238,7 @@ export class Search implements ComponentInterface {
                 this.search(inputValue);
             }
         }
+        this.changed.emit();
```
ammapspeople, line 241: Should this be moved inside the `if (!this.preventSearch)` statement? And I'm concerned about the return statement within that too. Do we risk not emitting when we should have?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -172,17 +177,48 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
+        setShowMyPositionOption(false);
+    }
+
+    /**
+     * Check if any of the fields have the 'USER_POSITION' selected.
+     * The 'USER_POSITION' option should only be available for one search field at a time.
+     */
+    function showMyPositionOptionButton(searchFieldIdentifier) {
+        if (originLocation?.id === 'USER_POSITION' && searchFieldIdentifier === searchFieldIdentifiers.TO) {
+            return setShowMyPositionOption(false);
```
ammapspeople, line 209: Do we need the `return`s?

andreeaceachir142: Actually not, removed the returns.

packages/components/src/components/search/search.tsx
andreeaceachir142: I moved it inside the statement. What would be the case when we would not emit when we should?

ammapspeople: From what I can see, we will not emit if the length of the input is less than two characters (due to the return). Should we emit in this case?

andreeaceachir142: Hmm, I think that is okay, isn't it? Because if there are less than two characters, we are emitting the shortInput event and we can use that if we need. Do you think we should emit the event in that case as well?

PR #257, opened 2023-08-22
Only load map provider specific components if needed

ammapspeople:
# What

- Implement lazy loading of map provider specific components.

# Why

- The map specific components import third party libraries, which may not be needed. This will make the initial bundle(s) smaller.

# How

- Use the `lazy` and `Suspense` parts of React to only load the MapboxMap or GoogleMapsMap if needed.


andreeaceachir142 (approved):
👏🏻

ammapspeople:
I'm closing this off for now because it's stale. We may decide to pick it up later.

PR #258, opened 2023-08-24
Add support for Google Maps Map ID prop

andreeaceachir142:
# What 

- Add support for Google Maps Map ID prop

# How 

- Pass the new `gmMapId` prop to the `GoogleMapsMap.jsx` inside the `mapViewOptions` object

### DIFF
--- a/packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
+++ b/packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
@@ -4,6 +4,7 @@ import mapsIndoorsInstanceState from '../../../atoms/mapsIndoorsInstanceState';
 import { Loader as GoogleMapsApiLoader } from '@googlemaps/js-api-loader';
 import gmApiKeyState from '../../../atoms/gmApiKeyState';
 import primaryColorState from '../../../atoms/primaryColorState';
+import gmMapIdState from '../../../atoms/gmMapIdState';
 
 /**
  * Takes care of instantiating a MapsIndoors Google Maps MapView.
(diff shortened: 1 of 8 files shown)


packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -39,7 +41,8 @@ function GoogleMapsMap({ onMapView, onPositionControl }) {
                 // Always set a center so the map so the bounds or center can be read from the start.
                 center: { lat: 0, lng: 0 },
                 // Set a large zoom so we prevent a "zoom 0 glitch" (showing the whole globe temporarily)
-                zoom: 21
+                zoom: 21, 
+                mapId: gmMapId
```
ammapspeople, line 45: Would it be more safe to exclude completely if not set? Does it work as expected when passing the default `null`?

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -19,6 +19,7 @@ import defaultLogo from "../../assets/logo.svg";
  * @param {array} [props.externalIDs] - Filter locations shown on the map based on the external IDs.
  * @param {string} [props.tileStyle] - Tile style name to change the interface of the map.
  * @param {number} [props.startZoomLevel] - The initial zoom level of the map.
+ * @param {string} [props.gmMapId] - The map ID associated with a specific map style or feature.
```
ammapspeople, line 22: We should probably specify that it is a Google Maps Map ID.

andreeaceachir142: Very good point. I'll specify that.

packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
ammapspeople: The atom has `null` as default. If that works, then no need to change anything 😄

andreeaceachir142: Yeah, it works 👍

ammapspeople (approved):
👏

PR #259, opened 2023-09-04
Fix my position URL parameter not working

andreeaceachir142:
# What 

- Fix my position URL parameter not working.

# How 

- Make the `watchPosition` a public method.


ammapspeople (approved):
Yes!

PR #264, opened 2023-09-11
Refactor usage of clear function on localStorage

andreeaceachir142:
# What 

- Refactor usage of `clear` function on localStorage

# How

- Use the `removeItem` function instead of the `clear` function


ammapspeople (approved):
Yeah... 🙃

PR #267, opened 2023-09-14
Implement search for Mapbox places 

andreeaceachir142:
# What 

- Implement search for Mapbox places 

# How 

- Use Search Box Api in order to suggest and retrieve Mapbox places, and calculate the route


packages/components/src/components/location-info/location-info.tsx
```diff
@@ -59,6 +59,11 @@ export class LocationInfo implements ComponentInterface {
             details.push(this.location.properties.subtitle);
         }
 
+        // Place formatter (For Mapbox usage)
+        if (this.location.properties.place_formatted) {
+            details.push(this.location.properties.place_formatted);
```
ammapspeople, line 64: There is no similar handling for Google? 🤔

packages/components/src/components/search/search.tsx
```diff
@@ -319,6 +337,46 @@ export class Search implements ComponentInterface {
...
+     * Get Mapbox Places results.
+     *
+     * @param {string} query
+     * @return {Promise<any>}
+     */
+    private getMapboxSearchResults(query: string): Promise<any> {
+        if (this.mapbox && this.mapboxAccessToken) {
+            if (query) {
+                return new Promise((resolve) => {
+                    const url = `https://api.mapbox.com/search/searchbox/v1/suggest?q=${query}&session_token=[GENERATED-UUID]&access_token=${this.mapboxAccessToken}`;
```
ammapspeople, line 350: I guess we need a proper `session_token` here.

packages/components/src/components/search/search.tsx
```diff
@@ -319,6 +337,46 @@ export class Search implements ComponentInterface {
...
+
+                    fetch(url)
+                        .then((response) => {
+                            return response.json();
+                        })
```
ammapspeople, line 355: Can be shortened to

```suggestion
                        .then(response => response.json())
```

packages/components/src/components/search/search.tsx
```diff
@@ -319,6 +337,46 @@ export class Search implements ComponentInterface {
...
+                                    placeId: result.mapbox_id,
+                                    name: result.name,
+                                    subtitle: result.place_formatted || '',
+                                    floor: 0
+                                }
+                            }));
+                            resolve(places);
+                        })
+                        .catch(err => {
+                            console.log('Error: ', err);
```
ammapspeople, line 371: Should we handle this better? Maybe just resolve with an empty array.

packages/map-template/src/components/Map/MapboxMap/MapboxPlacesHandler.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Handle Mapbox places and add geometry with type and coordinates.
+ *
+ * @param {Object} location
+ */
+function addMapboxPlaceGeometry(location, mapboxAccessToken) {
+    return new Promise((resolve, reject) => {
+        const url = `https://api.mapbox.com/search/searchbox/v1/retrieve/${location.id}?session_token=[GENERATED-UUID]&access_token=${mapboxAccessToken}`;
```
ammapspeople, line 8: I guess we need a proper `session_token` here.

packages/map-template/src/components/Map/MapboxMap/MapboxPlacesHandler.js
```diff
@@ -0,0 +1,31 @@
...
+            })
+            .then((result) => {
+                location.geometry = {
+                    type: 'Point',
+                    coordinates: [result.features[0].geometry.coordinates[0], result.features[0].geometry.coordinates[1]]
+                }
+                resolve(location);
+            })
+            .catch(err => {
+                console.log('Error: ', err);
```
ammapspeople, line 26: I guess we can reject here. Maybe.

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -114,14 +120,14 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
      */
     function locationClickHandler(location) {
         if (activeSearchField === searchFieldIdentifiers.TO) {
-            decorateLocation(location).then(location => {
-                setDestinationLocation(location);
-                toFieldRef.current.setDisplayText(location.properties.name);
+            decorateLocation(location).then(selectedLocation => {
```
ammapspeople, line 123: Why the change from `location` to `selectedLocation`?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -417,7 +425,7 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
                         {searchResults.map(location =>
                             <ListItemLocation
                                 key={location.id}
-                                icon={location.properties.type === 'google_places' ? googlePlacesIcon : undefined}
+                                icon={(location.properties.type === 'google_places' || location.properties.type === 'mapbox_places') ? googlePlacesIcon : undefined}
```
ammapspeople, line 428: Should we rename the `googlePlacesIcon` since it is now no longer exclusively used in a Google context?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: It was a bit confusing because we had `location` everywhere, and it made it more difficult to distinguish between them

packages/map-template/src/components/Map/MapboxMap/MapboxPlacesHandler.js
andreeaceachir142: Yes, that is in progress

packages/components/src/components/search/search.tsx
andreeaceachir142: Yes, this is in progress.

package.json
```diff
@@ -29,5 +29,8 @@
   ],
   "devDependencies": {
     "lerna": "^6.4.0"
+  },
+  "dependencies": {
```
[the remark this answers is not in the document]
andreeaceachir142: That is correct, I removed it ✅

packages/components/src/components/search/search.tsx
```diff
@@ -114,6 +119,11 @@ export class Search implements ComponentInterface {
      */
     @Prop() disabled: boolean = false;
 
+    /**
+     * The Mapbox Access Token for getting Mapbox Places.
```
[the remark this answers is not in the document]
andreeaceachir142: ✅

packages/components/src/components/location-info/location-info.tsx
andreeaceachir142: That's true, this check is not necessary anymore. Removed the added code ✅

packages/components/src/components/search/search.tsx
```diff
@@ -183,11 +192,19 @@ export class Search implements ComponentInterface {
...
+            this.makeGooglePlacesQuery(inputValue),
+            this.getMapboxSearchResults(inputValue),
         ])
             .then(results => {
                 this.lastRequested = inputValue;
-                this.pushResults(results[0].concat(results[1]));
+
+                if (this.google) {
+                    this.pushResults(results[0].concat(results[1]));
+                } else if (this.mapbox) {
```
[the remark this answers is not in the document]
andreeaceachir142: Yeah, that would be a good idea to have an extra check. ✅

packages/components/src/components/search/search.tsx
andreeaceachir142: I changed it now to `resolve([])`. Is this enough, or should we resolve the Promise?

packages/map-template/src/components/Map/MapboxMap/MapboxPlacesHandler.js
andreeaceachir142: ✅

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: Good catch, I renamed it to `locationMarkerIcon`. Do you have any other suggestion?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
ammapspeople: Could we call it `decoratedLocation` instead? I'm not really following the `selected` wording 🤔

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
ammapspeople: Maybe something like `externalLocationIcon` to indicate that it is used for only "external" locations? 🤔

andreeaceachir142: Good idea

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
andreeaceachir142: ✅

packages/components/src/components/search/search.tsx
andreeaceachir142: Do you think it should be refactored, or should we keep the resolve for now, and maybe some cases will be revealed after the testing of this feature?

ammapspeople: I suggest that we keep the resolve with the empty array for now.

ammapspeople:
Is there anyway to provide (and should we if so) a point so the results will be sorted by distance?

image

packages/map-template/package.json
```diff
@@ -7,7 +7,9 @@
     "dist/*.js",
     "dist/*.mjs"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "uuid": "^9.0.1"
```
ammapspeople, line 11: Should this be moved to devDependencies?

packages/map-template/src/atoms/sessionTokenState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const sessionTokenState = atom({
```
ammapspeople, line 3: Should we include Mapbox in the naming of this? It's only used for Mapbox.

packages/map-template/package.json
andreeaceachir142: ✅

packages/map-template/src/atoms/sessionTokenState.js
andreeaceachir142: ✅

packages/components/src/components/search/search.tsx
```diff
@@ -119,6 +124,16 @@ export class Search implements ComponentInterface {
      */
     @Prop() disabled: boolean = false;
 
+    /**
+     * The Mapbox Access Token used for getting Mapbox autocomplete suggestions.
+     */
+    @Prop() mapboxAccessToken: string;
```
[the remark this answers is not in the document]
andreeaceachir142: Nice, I refactored the component to use mapboxgl.accessToken.

andreeaceachir142:
> Is there anyway to provide (and should we if so) a point so the results will be sorted by distance?
> 
> image

Regarding this one, I will look into what we could do 🙏🏻 @ammapspeople

andreeaceachir142:
@ammapspeople 
I found that there is an optional parameter called `proximity` which could be used, by default it takes the IP address. https://docs.mapbox.com/api/search/search-box/#get-suggested-results Shouldn't that be enough?

ammapspeople:
> @ammapspeople I found that there is an optional parameter called `proximity` which could be used, by default it takes the IP address. https://docs.mapbox.com/api/search/search-box/#get-suggested-results Shouldn't that be enough?

Great find! It can also take a coordinate. That's probably better. But we need to talk to UX about what that coordinates should be: the current venue anchor? The user's position?

andreeaceachir142:
Implemented support for having user position as proximity parameter when getting directions @ammapspeople

packages/components/src/components/search/search.tsx
```diff
@@ -119,6 +125,16 @@ export class Search implements ComponentInterface {
      */
     @Prop() disabled: boolean = false;
 
+    /**
+     * The Mapbox Session Token used for getting Mapbox autocomplete suggestions.
+     */
+    @Prop() sessionToken: string;
```
ammapspeople, line 131: Clarification: Is this used outside this component? Can we generate the session token inside this component?

andreeaceachir142: Yes, this is used outside of the component as well. The reason for that is that we need to use the same session token in the search component `search.tsx` and also in the `MapboxPlacesHandler.js` file, when we retrieve the geometry. These two have to be the same, therefore we created an atom which stores the value, and we pass it down to the component.

ammapspeople: That's true. I was just hoping we could avoid the prop 😅

packages/components/src/components/search/search.tsx
```diff
@@ -188,11 +203,19 @@ export class Search implements ComponentInterface {
...
                 this.lastRequested = inputValue;
-                this.pushResults(results[0].concat(results[1]));
+
+                if (this.google) {
+                    this.pushResults(results[0].concat(results[1]));
+                } else if (this.mapbox && mapboxgl.accessToken) {
+                    this.pushResults(results[0].concat(results[2]));
+                } else {
+                    this.pushResults(results[0]);
+                }
```
ammapspeople, line 218: What happens if an insane developer provides both Google and Mapbox props? 😱

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
```diff
@@ -21,6 +23,7 @@ function MapboxMap({ onMapView, onPositionControl }) {
     const [hasPositionControl, setHasPositionControl] = useState(false);
     const mapsIndoorsInstance = useRecoilValue(mapsIndoorsInstanceState);
     const primaryColor = useRecoilValue(primaryColorState);
+    const [, setMapboxSessionToken] = useRecoilState(mapboxSessionTokenState);
```
ammapspeople, line 26: Why was it that we didn't want to store this in the browser's sessionStorage (https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)? Then the session will survive accross reloads, which may be what we want.

andreeaceachir142: Hmm, I remember we wanted to store the value in an atom because it made more sense to pass it down to the component. 🤔

packages/components/src/components/search/search.tsx
andreeaceachir142: Hmmm, right now if we provide both Mapbox and Google props as true on the search field, we get an error because the Google is not defined.
image

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
ammapspeople: It would be equally easy to read values from the sessionStorage. Atoms are like "lil session" 😝

packages/components/src/components/search/search.tsx
ammapspeople: Yeah, it's probably a scenario that will never happen. I guess we can dismiss it for now.

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
andreeaceachir142: Lil session 😅

andreeaceachir142: Do you think we should change it then?

ammapspeople: Yes, I think using sessionStorage to store a session token is more correct.

andreeaceachir142: Okay... good bye lil session

andreeaceachir142: The grown up session is taking over

andreeaceachir142:
Refactored the code to use `sessionStorage` to store the session token. @ammapspeople

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
```diff
@@ -38,6 +39,10 @@ function MapboxMap({ onMapView, onPositionControl }) {
         const externalDirectionsProvider = new window.mapsindoors.directions.MapboxProvider(mapboxAccessToken);
 
         onMapView(mapViewInstance, externalDirectionsProvider);
+
+        // Generate a UUIDv4 and set the Session Token for searching for Mapbox places.
+        const uuid = uuidv4();
+        sessionStorage.setItem('sessionToken', uuid);
```
ammapspeople, line 45: The name may be a little too generic. `mapboxPlacesSessionToken`?

andreeaceachir142: Renamed the key to `mapboxPlacesSessionToken`.

PR #272, opened 2023-10-03
Add support for pitch and bearing props

andreeaceachir142:
# What 

- Add support for pitch and bearing props

# How 

- Add new atoms for pitch and bearing states
- Adjust the my-position button to show the bearing of the compass correctly when the bearing prop is present 
- Check if the values of the pitch and bearing is a number, to avoid app crashing


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -98,6 +102,20 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
         }
     }, [filteredLocations, filteredLocationsByExternalIDs, mapsIndoorsInstance]);
 
+    /*
+     * React to changes in bearing and pitch props and set them on the map if mapsIndoorsInstance exists.
+     */
+    useEffect(() => {
+        if (mapsIndoorsInstance) {
+            if (pitch && !isNaN(parseInt(pitch))) {
```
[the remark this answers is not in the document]
andreeaceachir142: Done ✅

packages/map-template/src/atoms/bearingState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const bearingState = atom({
+    key: 'bearing',
+    default: null
```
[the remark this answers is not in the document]
andreeaceachir142: Good idea

packages/map-template/.github/README.md
```diff
@@ -76,6 +76,8 @@ When loading the MapsIndoors Map component for the first time, the map will resp
 |`startZoomLevel`|`number`|The initial zoom level of the map. |
 |`supportsUrlParameters`|`bool`|Indicates if the Map Template supports URL parameters. |
 |`gmMapId`|`string`|The Google Maps Map ID associated with a specific map style or feature. |
+|`pitch`|`pitch`|`number`|The pitch (Mapbox) or tilt (Google) value of the map. The value of the pitch can be between 0-90 degrees. |
```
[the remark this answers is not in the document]
andreeaceachir142: I specifically wrote the max value for Mapbox.

packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -62,7 +68,7 @@ function GoogleMapsMap({ onMapView, onPositionControl }) {
         if (mapsIndoorsInstance && mapView && google && !hasPositionControl) {
             const myPositionButtonElement = document.createElement('mi-my-position');
             myPositionButtonElement.mapsindoors = mapsIndoorsInstance;
-
+
```
ammapspeople, line 71:
```suggestion

```

packages/map-template/src/components/Map/Map.jsx
ammapspeople, line 110: A more strict check could be used instead of checking if `pitch` is "truthy".
If pitch is set to 0 and it previously was eg. 60, then this will not resolve to true. 

(the same applies to bearing).

andreeaceachir142: I read a bit more on this, and saw that it would be better practice to use `Number.isNaN` instead of `isNaN`. Do you think that could be enough? Or did I misunderstand your suggestion?

ammapspeople: As long as setting pitch to 0 will still trigger the call to tilt(), then do whatever works the best 😄

andreeaceachir142: Changed the code to use `Number.isNaN` instead, and it seems to be calling the function correctly

ammapspeople: But if `pitch` is 0, it will never reach `!Number.isNaN(parseInt(pitch))`, so it will never call `tilt()`. 

It is still making a loose evaluation against "truthyness" for the `pitch` variable.

ammapspeople: The problem is that the first part of the condition (`pitch`) will return false if the value is the number 0. 

We should not use that check. It's too "loose". And we should make sure that the condition as such returns what we want no matter if the value is `0`, `"0"`, `10`, `"10"`, `null`, `undefined`, `""`, `"aa"`, `[]`, `[0]`, `{}`, `{a: undefined}` etc.

andreeaceachir142: Refactored the conditions and the branch is now ready to be reviewed again.

ammapspeople (approved):
👏

PR #276, opened 2023-10-27
Location selection

andreeaceachir142:
# What 

- Integrate the new filter and search highlight into the Map Template (the SDK version will be changed when it is released, for now the one that is used is just a release candidate)
- When a category is selected, the locations belonging to that category are selected as well 
- When the user performs a search for a location which is in a different venue, the location is centered on the map, extracting the width / height of the modal / bottom sheet (200px for the bottom sheet) 
- The map zooms in to fit the selected location's bounds 
- The venue is updated accordingly
- The floor selector is updated accordingly 
- When exiting directions, regardless of which step it is, the floor selector is also updated correctly 
- When searching for locations belonging to the same venue, the floor selector updates accordingly 
- When hovering the locations in the search results list, the locations are highlighted on the map respectively, and the other way around
- When a location from a different venue is selected, the venue is updated but the fitVenue method is not called (that is achieved by having the atom `isLocationClickedState`)
- Add function for calculating the bounds of the location
- Add method that can be called when the floor is changed 

# How 

- Use the release candidate of the SDK 
- Create a new method in the floor selector component 
- Create a new function for calculating the bounds of the location
- Use the new methods provided by the SDK to filter and highlight locations on the map

Note! 
Regarding the `onLocationClicked()` function, I have tried to refactor it into a helper function, but that is using Recoil States and Values, which are not allowed in that case. Do you have any suggestions of how we could optimise this? @ammapspeople


ammapspeople:
@andreeaceachir142 A couple of thoughts regarding the onLocationClicked method:

1) Can we use a Recoil Selector? https://recoiljs.org/docs/basic-tutorial/selectors
2) Does it really belong in the BottomSheet and Sidebar components? Shouldn't it be in a more general component, for example the MapTemplate component?

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -95,9 +97,9 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
     useEffect(() => {
         if (mapsIndoorsInstance) {
             if (filteredLocations) {
-                mapsIndoorsInstance.filter(filteredLocations.map(location => location.id));
+                mapsIndoorsInstance.highlight(filteredLocations.map(location => location.id));
```
ammapspeople, line 100: If a customer is using another JS SDK version (we allow that - we only load the JS SDK if the `mapsindoors` namespace doesn't exist), this will break. Should we handle this better?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -293,6 +295,14 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setLogo(logo);
     }, [logo]);
 
+    useEffect(() => {
+        if (currentLocation) {
```
ammapspeople, line 299: If a customer is using another JS SDK version (we allow that - we only load the JS SDK if the mapsindoors namespace doesn't exist), this will probably throw an error. Should we handle this better?

packages/map-template/src/components/BottomSheet/BottomSheet.jsx
```diff
@@ -77,6 +84,35 @@ function BottomSheet({ directionsFromLocation, directionsToLocation, pushAppView
...
+        }
+
+        const currentFloor = mapsIndoorsInstance.getFloor();
+        const locationFloor = location.properties.floor;
+
+        // Set the floor to the one that the location belongs to.
+        if (locationFloor !== currentFloor) {
+            mapsIndoorsInstance.setFloor(locationFloor);
+            const floorSelectorElement = document.querySelector('mi-floor-selector');
+            floorSelectorElement.onFloorChanged(locationFloor)
```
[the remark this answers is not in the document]
andreeaceachir142: Refactored the floor selector component to listen to floor selector changes, and removed the `onFloorChanged` method. 
After investigating more into the issue, we have found out that the `display: none` on the floor selector element caused a problem when showing the active floor, and replacing it to `visibility: hidden` instead seems to be working. 
Still looking into the reason behind that not working...

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: Added a check if the method exists in the SDK. ✅

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Added a check if the method exists in the SDK. ✅

andreeaceachir142:
Refactored the `onLocationClicked` to happen in the `Search` component instead of the Sidebar and Bottom sheet separately. 
Feature ready to be reviewed again @ammapspeople

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -89,15 +91,25 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
     }, [venueName, venues, startZoomLevel]); // eslint-disable-line react-hooks/exhaustive-deps
     // We ignore eslint warnings about missing dependencies because mapsIndoorsInstance should never change runtime anyway.
 
-    /*
+    /**
```
ammapspeople, line 94: This is a function call, not a function definition, so it should not have a JSDoc block.

packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -189,11 +171,15 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped }) {
...
      * Reset the active steps and stop rendering directions.
+     * Set the floor according to the current location's floor.
      */
     function onDirectionsClosed() {
         resetSubsteps();
         stopRendering();
         onBack();
+
+        const currentFloor = mapsIndoorsInstance.getFloor();
+        mapsIndoorsInstance.setFloor(currentFloor);
```
ammapspeople, line 182: It is confusing why this is necessary: We read out the current floor and sets that floor to be the current floor 🤔

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -112,7 +114,7 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
 
             const miSdkApiTag = document.createElement('script');
             miSdkApiTag.setAttribute('type', 'text/javascript');
-            miSdkApiTag.setAttribute('src', 'https://app.mapsindoors.com/mapsindoors/js/sdk/4.24.7/mapsindoors-4.24.7.js.gz');
+            miSdkApiTag.setAttribute('src', 'https://folia-customer.s3.eu-west-1.amazonaws.com/mapsindoors/js/sdk/CandidateReleases/4.25.0-rc0/mapsindoors-4.25.0-rc0.js.gz');
```
ammapspeople, line 117: Remember to update this.

We can resolve this comment when done so we don't accidentally merge it 😄

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -139,6 +155,37 @@ function Search({ onLocationClick, onSetSize }) {
         }
     }
 
+    function onMouseEnter(location) {
```
ammapspeople, line 158: JSDoc

packages/map-template/src/components/Map/Map.jsx
andreeaceachir142: ✅

packages/map-template/src/components/Directions/Directions.jsx
andreeaceachir142: That is true, removed the unnecessary call for setting the floor. ✅

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: ✅

andreeaceachir142:
@ammapspeople ready to be reviewed again 😄

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -139,6 +156,44 @@ function Search({ onLocationClick, onSetSize }) {
...
+    /**
+     * Handle locations clicked on the map.
+     * 
+     * @param {object} location
+     */
+    function onLocationClicked(location) {
+        setCurrentLocation(location);
+
+        // Set the current venue to be the selected location venue.
+        if (location.properties.venueId !== currentVenueName) {
```
ammapspeople, line 177: This highlights a naming discrepancy. We are comparing an "ID" against a "name". Could we with low effort rename `currentVenueName` to `currentVenueId`?

packages/map-template/src/components/WebComponentWrappers/ListItemLocation/ListItemLocation.jsx
```diff
@@ -25,9 +26,19 @@ function ListItemLocation({ location, locationClicked, icon }) {
         current.icon = icon ? icon : mapsIndoorsInstance.getDisplayRule(location).icon;
 
         current.addEventListener('locationClicked', clickHandler);
-
-        return () => current.removeEventListener('locationClicked', clickHandler);
-    }, [location, locationClicked]);
+        current.addEventListener('mouseover', hoverHandler)
+        
+        if (isHovered) {
+            elementRef.current.classList.add('hovered')
```
ammapspeople, line 32: Can this be done in the JSX?

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: Done ✅

packages/map-template/src/components/WebComponentWrappers/ListItemLocation/ListItemLocation.jsx
andreeaceachir142: ✅

packages/map-template/src/components/Search/Search.jsx
ammapspeople: Ideally the recoil atom should be renamed as well, shouldn't it? (if that doesn't introduce other problems)

ammapspeople: We decided not to do that.

ammapspeople (approved):
✅

PR #279, opened 2023-11-08
Feature/build web component into one file

ammapspeople:
# What

- When building the Map Template as web component distribution files, bundle all files together in one. This is done to reduce the number of HTTP requests needed to run the Map Template as a Web Component.
- Made it possible to test the Map Template web component locally.

# How

- In the Rollup config, disable making chunks.
- Expose a new monorepo package, tests, which for now has a test HTML page for the locally built Map Template web component.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22827,7 +22827,7 @@
     },
     "packages/components": {
       "name": "@mapsindoors/components",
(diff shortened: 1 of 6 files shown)


andreeaceachir142 (approved):
Feature tested and approved ✅

PR #281, opened 2023-11-15
Implement kioskOriginLocationId prop

andreeaceachir142:
# What 

- Implement kioskOriginLocationId prop 
- Center the sidebar instead of having it placed in the left side
- Use the `fitBounds` function when having both `kioskOriginLocationId` and `locationId` props
- Restrict the kiosk to only be on desktop devices
- Create a new function `fitBoundsLocation` which is reused in three different places

# How 

- Refactor the `onLocationIdChanged` function to happen in a `useEffect`, and make sure the map is ready at the time the changes are happening 
- Handle both the `kioskOriginLocationId` and the `locationId` in the same `useEffect`
- Create helper functions for getting the desktop padding to the left and bottom, and the mobile padding to the bottom
- Do not allow location selection or clicking the location if it's the `kioskOriginLocationId` (meaning that it should not take the selection pin, and it should also not be selected to show the Location Details page)
- Make the kiosk icon double the size based on the display rule (UX Spec)
- Create a new function `fitBoundsLocation` which calculates the bbox of the location, takes a MapsIndoors instance and padding left and bottom, then it calls `fitBounds` method.

### DIFF
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -19,14 +19,16 @@ import tileStyleState from '../../atoms/tileStyleState';
 import startZoomLevelState from '../../atoms/startZoomLevelState';
 import positionControlState from '../../atoms/positionControlState';
 import locationIdState from '../../atoms/locationIdState';
-import setMapZoomLevel from "../../helpers/SetMapZoomLevel";
 import bearingState from '../../atoms/bearingState';
 import pitchState from '../../atoms/pitchState';
 import isLocationClickedState from "../../atoms/isLocationClickedState";
 import kioskOriginLocationIdState from "../../atoms/kioskOriginLocationIdState";
-import { calculateBounds } from "../../helpers/CalculateBounds";
 import currentKioskLocationState from "../../atoms/currentKioskLocationState";
-import getDesktopPadding from "../../helpers/GetDesktopPadding";
+import fitBoundsLocation from "../../helpers/fitBoundsLocation";
+import useMediaQuery from "../../hooks/useMediaQuery";
+import isMapReadyState from "../../atoms/isMapReadyState";
+import getMobilePaddingBottom from "../../helpers/GetMobilePaddingBottom";
+import getDesktopPaddingLeft from "../../helpers/GetDesktopPaddingLeft";
 
 const localStorageKeyForVenue = 'MI-MAP-TEMPLATE-LAST-VENUE';
 
@@ -66,6 +68,10 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
     const kioskOriginLocationId = useRecoilValue(kioskOriginLocationIdState);
     const [, setCurrentKioskLocation] = useRecoilState(currentKioskLocationState);
 
+    const isMapReady = useRecoilValue(isMapReadyState);
+
+    const isDesktop = useMediaQuery('(min-width: 992px)');
+
     useLiveData(apiKey);
 
     useEffect(() => {
@@ -86,7 +92,7 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
         if (mapsIndoorsInstance) {
             window.localStorage.removeItem(localStorageKeyForVenue);
             const venueToShow = getVenueToShow(venueName, venues);
-            if (venueToShow && !locationId && !isLocationClicked && !kioskOriginLocationId) {
+            if (venueToShow && !isLocationClicked) {
                 setVenue(venueToShow, mapsIndoorsInstance).then(() => {
                     onVenueChangedOnMap(venueToShow);
                 });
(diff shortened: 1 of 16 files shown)


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -320,12 +339,13 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
 
     /**
      * Handle the clicked location on the map.
-     * Set the current location if not in directions mode.
+     * Set the current location if not in directions mode,
+     * and if the clicked location is not the same as the kioskOriginLocationId.
```
ammapspeople, line 343: Should we motivate/describe why?

packages/map-template/src/components/Sidebar/Modal/Modal.jsx
```diff
@@ -41,7 +44,15 @@ function Modal({ children, isOpen }) {
         }
     }, [contentRef]);
 
-    return <div ref={modalRef} className={`modal ${isOpen ? 'modal--open' : ''} ${fullHeight ? 'modal--full' : ''} ${substeps ? 'modal--substeps' : ''}`}>
+    return <div ref={modalRef}
+        className={`modal ${isOpen ? 'modal--open' : ''} ${fullHeight ? 'modal--full' : ''} ${substeps ? 'modal--substeps' : ''}`}
+        style={{
```
ammapspeople, line 49: Could we add a BEM modifier instead?

ammapspeople (commented):
Do we need something added to READMEs and other documentation here and there?

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -264,6 +244,46 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
...
 
+    /*
+    * React on changes in the kioskOriginLocationId and locationId.
+    */
+    useEffect(() => {
+        if (isMapReady && mapsIndoorsInstance) {
+            if (kioskOriginLocationId && isDesktop) {
+                window.mapsindoors.services.LocationsService.getLocation(kioskOriginLocationId).then(kioskOriginLocation => {
+                    if (kioskOriginLocation) {
+                        // Replace icon with "you are here" icon
```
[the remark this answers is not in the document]
andreeaceachir142: Updated comments. ✅

packages/map-template/src/helpers/fitBoundsLocation.js
```diff
@@ -0,0 +1,13 @@
+import { calculateBounds } from "./CalculateBounds";
+
+/**
```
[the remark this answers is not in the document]
andreeaceachir142: Described params. ✅

packages/map-template/src/components/Map/Map.jsx
```diff
@@ -264,6 +244,46 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
...
+                    }
+                });
+            } else if (locationId) {
+                window.mapsindoors.services.LocationsService.getLocation(locationId).then(location => {
+                    if (location) {
+                        // Set the floor to the one that the location belongs to.
+                        const locationFloor = location.properties.floor;
+                        mapsIndoorsInstance.setFloor(locationFloor);
+
+                        fitBoundsLocation(location, mapsIndoorsInstance, 0, getDesktopPaddingLeft());
```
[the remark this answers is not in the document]
andreeaceachir142: Good point, implemented functionality for handling mobile as well. ✅

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Added a more detailed description. ✅

andreeaceachir142:
@ammapspeople Regarding this comment https://github.com/MapsPeople/web-ui/pull/281#pullrequestreview-1733840231, I was thinking to update the documentation everywhere at once, ideally at the end when we have an overview of all the combinations of parameters and so on. I could also add it now, for the `kioskOriginLocationId` only. Do you think that would be a better approach?

packages/map-template/src/components/Sidebar/Modal/Modal.jsx
andreeaceachir142: That's a good idea, I added a BEM modifier instead.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
ammapspeople: 👏

packages/map-template/src/components/Sidebar/Modal/Modal.jsx
ammapspeople: 🙌

andreeaceachir142:
Ready to be reviewed again @ammapspeople

ammapspeople (approved):
👏

PR #282, opened 2023-11-17
Use map provider tokens if set in the app config

ammapspeople:
# What

- Make is possible to use Mapbox Access Token and Google Maps API key from the solution's App Config.

# How

- Set app config as `useState` early when loading data.
- Use access token / API key from the app config if there is none set otherwise.
- Remove the `useRef` for app config since that is no longer needed.


packages/map-template/README.md
```diff
@@ -136,21 +136,21 @@ The Map Template supports using query parameters for all the properties provided
...
+8. `directionsTo` - Used like this `directionsTo=yourDestinationLocationId` when having a location ID, or like this `directionsTo=USER_POSITION` when having the user location.
+9. `externalIDs` - Used like this `externalIDs=0.0.1,0.0.2,0.0.3`. **Note!** You need to provide a list of comma separated values, without any spaces between the comma and the value. This will further be converted into an array of external IDs. Because of the way browsers work, you **cannot** use External IDs with the `,`, `&`, `#` and `+`, character in them, as they are interpreted by the browser in a particular way.
+10. `tileStyle` - Used like this `tileStyle=yourTileStyleName`. If no tile style is provided, the app will show the default tile style.
+11. `mapboxAccessToken` - Used like this `mapboxAccessToken=yourMapboxAccessToken`. If both the `mapboxAccessToken` and the `gmApiKey` are present, the app will load a Mapbox map.
+12. `gmApiKey` - Used like this `gmApiKey=yourGmApiKey`. If both the `mapboxAccessToken` and the `gmApiKey` are present, the app will load a Mapbox map.
```
andreeaceachir142, line 150: Should we update the docs here as well? @ammapspeople

andreeaceachir142:
@ammapspeople we should update the README under the .github folder as well `packages/map-template/.github`

ammapspeople:
@andreeaceachir142 I hope I updated all the docs now 🙏

andreeaceachir142:
Approved 🎉

PR #284, opened 2023-11-17
Add support for having one external ID 

andreeaceachir142:
# What 

- Add support for having one external ID to behave like the `locationID` parameter

# How 

- Add condition when having one external ID to interpret that as a `locationID` parameter


ammapspeople (approved):
We probably need to think about avoiding the duplicate code in the Sidebar and BottomSheet sometime in the future. But that is a major refactor, so: approved 😄

andreeaceachir142:
As agreed, I have created a ticket for refactoring the useEffect inside the Sidebar and Bottom sheet components. @ammapspeople

PR #286, opened 2023-11-20
Add missing stable tag for docs on NPM

andreeaceachir142:
# What

- Add missing `stable` tag for docs on NPM

### DIFF
--- a/packages/map-template/README.md
+++ b/packages/map-template/README.md
@@ -91,7 +91,7 @@ Install the package:
 Use the `MapsIndoorsMap` component in a React component:
 
 ```javascript
-import MapsIndoorsMap from '@mapsindoors/map-template/dist/mapsindoors-react.es.js';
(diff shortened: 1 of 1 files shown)


packages/map-template/README.md
```diff
@@ -11,7 +11,7 @@ Install the package:
 In your script:
 
 ```javascript
-import MapsIndoorsMap from '@mapsindoors/map-template/dist/mapsindoors-webcomponent.es.js';
+import MapsIndoorsMap from '@mapsindoors/map-template@stable/dist/mapsindoors-webcomponent.es.js';
```
ammapspeople, line 14: It is not possible to use versions or tags when importing from an already installed package. 

Using a tag should be done at install time.

andreeaceachir142: Oh, I wasn't aware of that. I just saw a misalignment with the docs site, and wanted to make sure all the content is correct. I will close this PR.

PR #287, opened 2023-11-20
Show directions immediately after selecting a location

andreeaceachir142:
# What 

- Show directions immediately after selecting a location

# How 

- Implement functionality for calculating the route when selecting a location (destination location)


packages/map-template/src/atoms/currentKioskLocationState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const currentKioskLocationState = atom({
```
ammapspeople, line 3: Does this contain the whole Location or just the ID? Can it be merged with the `kioskOriginLocationIdState`?

packages/map-template/src/components/LocationDetails/LocationDetails.jsx
```diff
@@ -52,8 +73,46 @@ function LocationDetails({ onBack, onStartWayfinding, onSetSize, snapPointSwiped
...
+        if (currentKioskLocation) {
+            setOriginLocation(currentKioskLocation)
+        }
+    }, [location, mapsIndoorsInstance, currentKioskLocation]);
+
+    /*
+     * When both origin location and destination location are selected, and have geometry, 
+     * call the MapsIndoors SDK to get information about the route.
+     */
+    useEffect(() => {
```
ammapspeople, line 89: I think we should put some effort into not having duplicate code. This is more or less the same as we have in the Wayfinding component. Can we extract it into a reusable component? Or even a custom hook?

packages/map-template/src/atoms/currentKioskLocationState.js
andreeaceachir142: The `currentKioskLocationState` contains the whole location, which we need for getting directions. So we need that.

ammapspeople: Can we then use this one instead of `kioskOriginLocationIdState`?

andreeaceachir142: Hmm, I'm not sure we can, because we need to first listen to the changes in the kioskOriginLocationId prop, and then based on that id, we get the location and then set the setCurrentKioskLocation on line 262.

ammapspeople: Yes. We already listen for changes in the `kioskOriginLocationId` prop. Reacting to that, we can read the location and store it in an atom. And use that atom wherever we now use the kioskLocationId atom. That way we will only have one atom to store kiosk origin info instead of two. Wouldn't that be cleaner?

andreeaceachir142: Removed the `travelMode` and `accessibilityOn` dependencies from the useEffect, as they are passed through props.

ammapspeople: Does that fix it? 🙏

ammapspeople (approved):
👏 🥳 🎉

andreeaceachir142:
Fixed the merge conflicts occurring from main 🙏🏻 @ammapspeople

ammapspeople:
Reapproved ✅

PR #288, opened 2023-11-23
Feature/load map provider based on solution module

ammapspeople:
# What

- Use Solution modules to determine what map type to load.

# How

- Read the Solution information as early as possible. 
- Refactor the decision logic to choose map provider.
- Implement a Notification component wrapping a `<mi-notification`> component. This is used to show an error message in case no map type could be determined.
- Refactor the SetMapZoomLevel helper into a custom React hook.


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -74,15 +79,61 @@ function Map({ onLocationClick, onVenueChangedOnMap }) {
...
+        false                    ✅                         ✅                       ❌                      Mapbox
+        false                    ✅                         ❌                       ✅                      Google Maps
+        false                    ✅                         ❌                       ❌                      Google Maps
+        false                    ❌                         ✅                       ✅                      Mapbox
+        false                    ❌                         ✅                       ❌                      Mapbox
+        false                    ❌                         ❌                       ✅                      Google Maps
+        false                    ❌                         ❌                       ❌                      Google Maps
+         */
+
+        let mapTypeToUse;
```
andreeaceachir142, line 110: Could we use a `useState` here instead?

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -60,6 +62,7 @@ function MapsIndoorsMap(props) {
         const appUserRolesQueryParameter = queryStringParams.get('appUserRoles')?.split(',');
         const externalIDsQueryParameter = queryStringParams.get('externalIDs')?.split(',');
         const gmMapIdQueryParameter = queryStringParams.get('gmMapId');
+        const useMapProviderModule = queryStringParams.get('useMapProviderModule');
```
andreeaceachir142, line 65: Maybe use the convention above `useMapProviderModuleParameter` here. 
I can see that on this branch the `kioskOriginLocationId` is not following the convention either, but that has been fixed in another branch 😄

packages/map-template/src/components/Notification/Notification.jsx
```diff
@@ -0,0 +1,29 @@
...
+            elementRef.current.clearAll();
+        }
+    }, [notificationMessage]);
+
+    return (<>
+        <mi-notification ref={elementRef} position="top-center"></mi-notification>
+    </>)
+}
+
+export default Notification;
```
andreeaceachir142, line 29: Does this component belong to the `WebComponentWrappers` folder instead? I'm thinking that it follows the same pattern as all the other components more or less.

packages/map-template/src/components/Notification/Notification.jsx
```diff
@@ -0,0 +1,29 @@
...
+        if (notificationMessage) {
+            elementRef.current.push(notificationMessage.text, notificationMessage.type, true);
+        } else {
+            elementRef.current.clearAll();
+        }
+    }, [notificationMessage]);
+
+    return (<>
+        <mi-notification ref={elementRef} position="top-center"></mi-notification>
+    </>)
```
andreeaceachir142, line 26: Do we need the `(<>` anymore here, since we are only rendering the `mi-notification` element?

packages/map-template/src/components/Notification/Notification.scss
```diff
@@ -0,0 +1,4 @@
+.notification {
+    z-index: 10000000;
+    min-width: 320px;
```
andreeaceachir142, line 3: Do we need to have a max width?

packages/map-template/src/components/Notification/Notification.scss
```diff
@@ -0,0 +1,4 @@
+.notification {
+    z-index: 10000000;
```
andreeaceachir142, line 2: Is the z-index set a level higher than the splash screen in order to be the most visible element?

packages/map-template/src/components/Notification/Notification.jsx
ammapspeople: No we dont. Fixed.

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -79,6 +82,7 @@ function MapsIndoorsMap(props) {
             appUserRoles: props.supportsUrlParameters && appUserRolesQueryParameter ? appUserRolesQueryParameter : props.appUserRoles,
             externalIDs: props.supportsUrlParameters && externalIDsQueryParameter ? externalIDsQueryParameter : props.externalIDs,
             gmMapId: props.supportsUrlParameters && gmMapIdQueryParameter ? gmMapIdQueryParameter : props.gmMapId,
+            useMapProviderModule: props.supportsUrlParameters && useMapProviderModuleParameter ? useMapProviderModuleParameter : (props.useMapProviderModule || defaultProps.useMapProviderModule),
             kioskOriginLocationId: props.supportsUrlParameters && kioskOriginLocationId ? kioskOriginLocationId : props.kioskOriginLocationId,
         });
     }, [props]);
```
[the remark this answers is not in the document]
ammapspeople: I did it!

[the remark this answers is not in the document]
andreeaceachir142: 🥹👏🏻

andreeaceachir142 (approved):

andreeaceachir142:
👏🏻

andreeaceachir142:
@ammapspeople We should remember to update all the docs places with the new prop for enabling the solution module 🙏🏻

andreeaceachir142 (approved):

andreeaceachir142 (approved):

PR #290, opened 2023-11-24
Implement useKeyboard prop

andreeaceachir142:
# What 

- Implement prop `showKeyboard` for showing the onscreen keyboard
- Upgrade version of package used for the `mi-keyboard` component 

# How 

- Add prop for showing onscreen keyboard 
- Restrict the keyboard to only be shown in a kiosk context, on desktop
- Create a new method on the `mi-search` component that returns the input field, which is needed in order to pass it as a prop to the keyboard component 
- Create a wrapper component around the `mi-keyboard` component


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -326,6 +329,19 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         }
     }, [kioskOriginLocationId, mapsindoorsSDKAvailable]);
 
+    /*
+     * React on changes to the showKeyboard prop.
+     */
+    useEffect(() => {
+        if (mapsindoorsSDKAvailable) {
+            if (showKeyboard === 'true' || showKeyboard === true) {
```
ammapspeople, line 337: I think this should be handled where we handle query params and props (MapsIndoorsMap component). It should be converted to an actual boolean, so we don't have to make this guard in other components.

ammapspeople: Then we can also make this simpler, `setShowKeyboard(showKeyboard);`.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -154,6 +163,10 @@ function Search({ onSetSize }) {
         } else {
             searchFieldRef.current.focusInput();
         }
+
+        if (showKeyboard) {
```
ammapspeople, line 167: Is this necessary? Couldn't we just use the Recoil value in the JSX?

packages/map-template/src/components/WebComponentWrappers/Keyboard/Keyboard.jsx
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useRef } from 'react';
+
+/**
+ * React wrapper around the custom element <mi-keyboard>.
+ *
+ * @param {object} props
+ * @param {object} props.searchInputElement
```
ammapspeople, line 7: Write a few words about what this is.

ammapspeople:
Will it be added to the Wayfinding component later, or should that be done in this PR?

ammapspeople:
Is the keyboard detecting browser language in order to show eg. Danish or German layout keyboards?

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: There was a missing function call here for the `getInputField` on the search ref component 😄

andreeaceachir142: But I guess we could get rid of the useState `isKeyboardVisible` and just use the Recoil value

packages/map-template/src/components/WebComponentWrappers/Keyboard/Keyboard.jsx
andreeaceachir142: ✅

ammapspeople: Should it be a HTMLInputElement instead of an object?

andreeaceachir142: Done ✅

andreeaceachir142:
> Is the keyboard detecting browser language in order to show eg. Danish or German layout keyboards?

I don't think it does, but as agreed with Mark, when we get the translations in the Map Template, we'll have to hook it up with the keyboard as well.

andreeaceachir142:
> Will it be added to the Wayfinding component later, or should that be done in this PR?

As discussed 2 days ago, the keyboard will not be needed on the Wayfinding, since it will only be present in a kiosk context which never goes to Wayfinding.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Done ✅

andreeaceachir142:
Ready for review again 🙏🏻 @ammapspeople

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -80,6 +83,7 @@ function MapsIndoorsMap(props) {
             externalIDs: props.supportsUrlParameters && externalIDsQueryParameter ? externalIDsQueryParameter : props.externalIDs,
             gmMapId: props.supportsUrlParameters && gmMapIdQueryParameter ? gmMapIdQueryParameter : props.gmMapId,
             kioskOriginLocationId: props.supportsUrlParameters && kioskOriginLocationId ? kioskOriginLocationId : props.kioskOriginLocationId,
+            showKeyboard: props.supportsUrlParameters && showKeyboardQueryParameter ? showKeyboardQueryParameter :(props.showKeyboard || defaultProps.showKeyboard),
```
ammapspeople, line 86:
```suggestion
            showKeyboard: props.supportsUrlParameters && showKeyboardQueryParameter ? showKeyboardQueryParameter : (props.showKeyboard || defaultProps.showKeyboard),
```

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -91,3 +95,15 @@ function MapsIndoorsMap(props) {
...
+
+
+/**
+ * Convert query parameter value (which is always a string) into a boolean.
+ * It will only accept the string 'true' as a boolean true. Anything else will return false.
+ *
+ * @param {string} queryParameterValue
+ * @return {boolean}
+ */
+function getBooleanQueryParameter(queryParameterValue) {
```
ammapspeople, line 107: Looks slightly familiar 😅

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -31,9 +34,13 @@ function Search({ onSetSize }) {
     /** Referencing the search field */
     const searchFieldRef = useRef();
 
+    /** Referencing the keyboard element */
+    const keyboardRef = useRef();
+
     const [searchDisabled, setSearchDisabled] = useState(true);
     const [searchResults, setSearchResults] = useState([]);
     const categories = useRecoilValue(categoriesState);
+    const showKeyboard = useRecoilValue(showKeyboardState);
```
ammapspeople, line 43: showKeyboard can also mean that it is currently shown. Should the atom (and even the prop) be renamed into something like `useKeyboard`?

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -91,3 +95,15 @@ function MapsIndoorsMap(props) {
...
+    return queryParameterValue === 'true' ? true : false;
```
ammapspeople, line 108: it can even be shortened to

```suggestion
    return queryParameterValue === 'true';
```

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: Good idea, I renamed the atom and the prop.

ammapspeople: 👏

andreeaceachir142 (commented):
Approved ✅

packages/components/src/components/keyboard/keyboard.tsx
```diff
@@ -60,13 +60,26 @@ export class Keyboard implements ComponentInterface {
...
+    }
+
     private simpleKeyboard: SimpleKeyboard;
     private inputElements = new Set<HTMLInputElement>();
 
     componentDidLoad(): void {
         this.simpleKeyboard = new SimpleKeyboard({
             onChange: (input: string) => {
                 if (this.inputElement) {
+                    console.log('running', input);
```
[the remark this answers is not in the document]
andreeaceachir142: ✅

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -265,6 +283,23 @@ function Search({ onSetSize }) {
         }
     });
 
+    useEffect(() => {
```
ammapspeople, line 286: Add a comment explaining what we are doing here, and why.

packages/map-template/src/components/WebComponentWrappers/Keyboard/Keyboard.jsx
```diff
@@ -0,0 +1,38 @@
...
+ * @param {HTMLInputElement} props.searchInputElement
+ *
+ */
+const Keyboard = forwardRef(({ searchInputElement }, ref) => {
+    const elementRef = useRef();
+
+    const language = useRecoilValue(languageState);
+
+    /**
+     * Methods that can be triggered on the mi-search element.
```
ammapspeople, line 19: I don't think we should mention mi-search here 😄

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: ✅

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -265,6 +283,23 @@ function Search({ onSetSize }) {
...
+        if (useKeyboard) {
+            const onClick = (event) => {
+                if (event.target.tagName.toUpperCase() === 'DIV' || event.target.tagName.toUpperCase() === 'MI-SEARCH' || event.target.tagName.toUpperCase() === 'INPUT') {
```
ammapspeople, line 293: I think the DIV tagName is a bit too generic. I can trigger the keyboard by clicking on any div element, here the div that acts as container for Mapbox info:

https://github.com/MapsPeople/web-ui/assets/50314165/d322d55c-dfd0-46c7-a786-f11128dcdb00

andreeaceachir142: Hmmm, that is true. Is there a way we can target the element's class in this case? We need to target the clicks on the div element with the certain class here 🤔
image

andreeaceachir142: Nevermind, I found a way 😄

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -265,6 +283,32 @@ function Search({ onSetSize }) {
...
 
+    /*
+     * When useKeyboard parameter is present, add click event listener which determines when the keyboard should be shown or not. 
+     */
+    useEffect(() => {
+        if (useKeyboard) {
+            const onClick = (event) => {
+                // Detect the className of the keyboard buttons, the space and the backspace
+                // If the user clicks on the keyboard or the search fields, the keyboard should stay visible 
+                if (event.target.className === 'hg-button hg-standardBtn' ||
```
ammapspeople, line 294: This is very specific to the internals of the keyboard component, which may change in a future version.

Could we instead try something with the `closest()` method (https://developer.mozilla.org/en-US/docs/Web/API/Element/closest), for example `event.target.closest('mi-keyboard')` to catch the clicks on the keyboard?

andreeaceachir142: Very nice suggestion! I think it works well 👌🏻

packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
```diff
@@ -97,6 +103,7 @@ const SearchField = forwardRef(({ placeholder, mapsindoors, results, clicked, cl
     }, [placeholder, mapsindoors, results, clicked, cleared, google, mapbox, changed]);
 
     return <mi-search ref={elementRef}
+        id-attribute='id'
```
ammapspeople, line 106: Was this needed, or was that part of some debugging? I can't remember.

andreeaceachir142: I think it was needed to fix the warning

ammapspeople:
Re-approved ✅

PR #294, opened 2023-12-06
Implement dialog with QR code in the Kiosk

andreeaceachir142:
# What 

- Implement dialog with QR code in the directions 
- Add logic for handling QR code URL by reading the current URL where the Map Template exists (in a kiosk context indicated by the `kioskOriginLocationId` prop), then check for `mapboxAccessToken` and `gmApiKey` URL parameters
- Handle the `primaryColor`, `logo` and `apiKey` props and URL parameters in order to construct the QR code URL 
- Add `directionsTo` and `directionsFrom` URL parameters when constructing the QR code URL 
- Set the `kioskOriginLocationId` to be the `directionsFrom` in the QR code URL 

# How 

- Install `qrcode` NPM package 
- Add button on the directions that open up QR code dialog 
- Create new page for the QR code dialog
- Create an atom which indicates if the QR code should be visible 
- Add logic for handling the QR code URL
- Add support for having QR code button only in Kiosk mode


packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                }
+            }
+
+            // Handle query parameters for the gmApiKey and mapboxAccessToken
+            handleQueryParams('gmApiKey');
+            handleQueryParams('mapboxAccessToken');
+
+            // Handle query parameters and props for apiKey, primaryColor, logo 
+            handleQueryParamsAndProps('apiKey', apiKeyProp);
```
ammapspeople, line 76: Can you explain why these three props should be handled differently from the two above?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+            const options = {
+                errorCorrectionLevel: 'L',
+                margin: 0,
+                width: '225'
+            };
+
+            QRCode.toDataURL(QRCodeURL, options)
+                .then((dataUrl) => {
+                    document.getElementById('qr').src = dataUrl;
```
ammapspeople, line 94: Can we do this with JSX?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,117 @@
...
+                }).catch(() => {
+                    console.log('error')
```
ammapspeople, line 96: Nah 😄

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+    function closeDialog() {
+        setShowQRCodeDialog(false);
+    }
+    return (<>
+        <div className="background"></div>
+        <div className="qr-code">
+            <img id='qr' alt="QR Code" className="qr-code__image" />
+            <p>Scan the QR code to see the route on your phone</p>
+            <button className="qr-code__button" style={{ background: primaryColorProp }} onClick={() => closeDialog()}>DONE</button>
```
ammapspeople, line 111: Use CSS to transform the text to uppercase.

packages/map-template/package.json
```diff
@@ -8,7 +8,8 @@
     "dist/*.mjs"
   ],
   "dependencies": {
-    "@turf/turf": "^6.5.0"
+    "@turf/turf": "^6.5.0",
+    "qrcode": "^1.5.3"
```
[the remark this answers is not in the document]
andreeaceachir142: ✅

packages/map-template/src/atoms/showQRCodeDialogState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const showQRCodeDialogState = atom({
```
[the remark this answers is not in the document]
andreeaceachir142: Yes, I think that could be a good idea, the list of atoms seems to be getting longer and longer. 
Should I do this as part of this PR, or should I create a ticket for handling that?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,117 @@
...
+    const apiKeyProp = useRecoilValue(apiKeyState);
+    const logoProp = useRecoilValue(logoState);
+
+    const directionsFrom = useRecoilValue(kioskLocationState);
+    const directionsTo = useRecoilValue(currentLocationState);
+
+    useEffect(() => {
+        if (directionsFrom && directionsTo) {
+            // The current URL when the user opens the QR code dialog
+            let currentUrl = window.location.origin;
```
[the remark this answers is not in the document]
andreeaceachir142: ✅

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: Because in the case of apiKey, primaryColor and logo we need to have the else statement which checks for the existence of the prop. 
My thought was that in the case of gmApiKey and mapboxAccessToken we don't need to do that, as one of them will always be present, therefore we only need to check the query params. 
Do you think it would make sense to handle all the props the same as we do for the apiKey, primaryColor and logo?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: ✅

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: Hah, I must have missed that. 😅 What should happen in the `catch`?

andreeaceachir142: Do you have any suggestion on how we should handle that?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: Hmm, the `document.getElementById` part?

andreeaceachir142: Yes, it is possible with a `useRef`  🎉

ammapspeople: 🥇

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
ammapspeople: The question is if we should handle it at all. What can go wrong? Have you seen it failing?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,114 @@
...
+                });
+        }
+
+    }, [directionsFrom, directionsTo])
+
+    return (<>
+        <div className="background"></div>
+        <div className="qr-code">
+            <img alt="QR Code" className="qr-code__image" ref={elementRef} />
+            <p>Scan the QR code to see the route on your phone</p>
```
ammapspeople, line 106: We will probably need to get this in French, German and Danish. Let's see what goes into main first 😆

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
ammapspeople: Yeah, wouldn't it be more simple to handle it in the same way?

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: Hmm no actually. I will remove the `catch` code block.

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: 💨🏃🏻 Run!

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,117 @@
...
+                    if (prop === primaryColorProp) {
+                        newParams.append(queryParam, primaryColorProp.replace("#", ""))
+                    } else {
+                        newParams.append(queryParam, prop)
+                    }
+                }
+            }
+
+            // Handle query parameters for the gmApiKey and mapboxAccessToken
+            handleQueryParams('gmApiKey');
```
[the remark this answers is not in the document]
andreeaceachir142: ✅

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: Yes, that's true. Refactored the code to respect that ✅

packages/map-template/src/atoms/showQRCodeDialogState.js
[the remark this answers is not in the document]
ammapspeople: I guess we need to hurry with this, so create a new ticket.

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,95 @@
...
+                        if (prop === primaryColorProp) {
+                            newParams.append(queryParam, primaryColorProp.replace("#", ""))
+                        } else {
+                            newParams.append(queryParam, prop)
+                        }
+                    }
+                });
+
+            // Get the string with all the final query parameters
+            const finalParams = newParams.toString()
```
ammapspeople, line 65:
```suggestion
            const finalParams = newParams.toString();
```

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+            // Construct the QR code URL
+            let QRCodeURL = `${targetUrl}/?${finalParams}&directionsFrom=${directionsFrom.id}&directionsTo=${directionsTo.id}`
```
ammapspeople, line 68:
```suggestion
            let QRCodeURL = `${targetUrl}/?${finalParams}&directionsFrom=${directionsFrom.id}&directionsTo=${directionsTo.id}`;
```

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,95 @@
...
+                width: '225'
+            };
+
+            QRCode.toDataURL(QRCodeURL, options)
+                .then((dataUrl) => {
+                    elementRef.current.src = dataUrl;
+                });
+        }
+
+    }, [directionsFrom, directionsTo])
```
ammapspeople, line 82:
```suggestion
    }, [directionsFrom, directionsTo]);
```

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,95 @@
...
+            ['apiKey', apiKeyProp],
+            ['primaryColor', primaryColorProp],
+            ['logo', logoProp]]
+                .forEach(([queryParam, prop]) => {
+                    if (currentParams.has(queryParam)) {
+                        const queryParameter = currentParams.get(queryParam);
+                        newParams.append(queryParam, queryParameter);
+                    } else if (prop) {
+                        if (prop === primaryColorProp) {
+                            newParams.append(queryParam, primaryColorProp.replace("#", ""))
```
ammapspeople, line 57:
```suggestion
                            newParams.append(queryParam, primaryColorProp.replace("#", ""));
```

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,95 @@
...
+                        } else {
+                            newParams.append(queryParam, prop)
```
ammapspeople, line 59:
```suggestion
                            newParams.append(queryParam, prop);
```

Not a fan of semicolons anymore? 😆

packages/map-template/src/atoms/showQRCodeDialogState.js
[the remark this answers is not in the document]
andreeaceachir142: Done ✅

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
andreeaceachir142: It was a test for you 🙄 To see if you pay attention

ammapspeople (approved):
🆗 👍

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -0,0 +1,95 @@
...
+    const directionsFrom = useRecoilValue(kioskLocationState);
+    const directionsTo = useRecoilValue(currentLocationState);
+
+    useEffect(() => {
+        if (directionsFrom && directionsTo) {
+            // The target URL when the user opens the QR code dialog
+            let targetUrl = window.location.origin;
+
+            // The interface for the existing URL search params
+            const currentParams = new URLSearchParams(window.location.search);
```
ammapspeople, line 37: Just came to mind: Should we check for supportsUrlParameters here? We risk clashing with similarly named query parameters in case the customer has their own.

andreeaceachir142: Hmmm, sure, we could check for that prop here, but by default we assume that the `supportsUrlParameters` is true. So that would mitigate the risk of clashing in case we the customer user **props only** and has `supportsUrlParameters` set to false.
But if the customer has something like `primaryColor` or `logo`, and if we have `supportsUrlParameters` set to true, it would would clash anyway, right? @ammapspeople

ammapspeople: Imagine the case where

1. The customer has `supportsUrlParameters` to false (which is the default)
2. Have their own URL parameter scheme in place. Some of these are named `logo` and `primaryColor`

What will happen to the QR code URL?

andreeaceachir142: That's a good point 💡 I will add a check for `supportsUrlParameters`

andreeaceachir142: Ready for review again 😄 @ammapspeople

ammapspeople: @andreeaceachir142 I maintain my approval 😄

andreeaceachir142: 🎉

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
ammapspeople: I guess you won 😭

andreeaceachir142: Hahah 😂 Oops

PR #296, opened 2023-12-11
Feature/support multiple languages

ammapspeople:
# What

- Make it possible to run the Map Template in four different languages, either by auto-detection or enforcing a language with a new `language` prop or query parameter.

# How

- Install new dependencies for managing i18n: `i18next` and `i18next-react`
- Setup language files in an `i18n` folder
- Use the `useTranslation` hook to add translation of all hardcoded text strings
- Added translations for Danish and German.
- Added `level` prop on the `list-item-location` web component.
- Repopulate state when the MapsIndoors JS SDK emits the `update_completed` event, which indicates that new (translated) Locations data has been read.


packages/components/src/components/list-item-location/readme.md
```diff
@@ -86,6 +86,7 @@ A `listItemDidRender` event is emitted from the `<mi-list-item-location>` elemen
 | `icon`           | `icon`             |                                  | `string`                                   | `undefined` |
 | `iconBadge`      | `icon-badge`       |                                  | `string`                                   | `undefined` |
 | `iconBadgeValue` | `icon-badge-value` |                                  | `string`                                   | `undefined` |
+| `level`          | `level`            |                                  | `string`                                   | `'Level'`   |
```
andreeaceachir142, line 89: Can we just use `Level` without the '' ?

ammapspeople: It signifies that the default value is the string `'Level'`.

andreeaceachir142: Alright 😄

andreeaceachir142 (approved):
Great job 👏🏻

andreeaceachir142 (approved):

PR #297, opened 2023-12-14
Add support for startZoomLevel, pitch and bearing working in combination with other props

andreeaceachir142:
# What 

- Add support for `startZoomLevel`, `pitch` and `bearing` to be working in combination with locationId and kioskOriginLocationId

# How

- In the `fitBoundsLocation()` function, when calling the `fitBounds()` from the SDK, handle the presence of startZoomLevel, pitch and bearing properties.
- Add the additional 3 props when calling `fitBoundsLocation()` when having the `locationId` and `kioskOriginLocationId` props


ammapspeople (approved):
No comments. Great job 👍

PR #303, opened 2023-12-29
Reset after timeout

ammapspeople:
# What

- Make it possible to instruct the Map Template to reset map position and UI to initial state after x seconds of user inactivity.

# How

- A large refactor of how map position is determined, most notable by a new custom hook, `useMapBoundsDeterminer`, which handles setting the map bounds based on venue, locationId and kioskOriginLocationId props.
    - Also a refactor of the use of `isMapReady` and a new `isMapPositionKnown` variable.
- Re-introduced the kioskOriginLocationId state since it it is needed to avoid timing issues.
- No longer using the MapsIndoors SDK's fitVenue method, since animation etc. is not controllable.
- A venue state (Recoil atom) that holds the value of the venue prop. This is needed in order to reset to the set venue, not just the current venue.
- The methods for getting padding are "promise-ified" since it may be called in situations where the inspected elements are not there yet.
- A new custom hook, `useInactive` which will measure and report user inactivity.
- A new custom hook, `useReset` that resets carefully selected Recoil Atoms.


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -49,35 +36,26 @@ let _tileStyle;
  * @param {boolean} props.useMapProviderModule - If you want to use the Map Provider set on your solution in the MapsIndoors CMS, set this to true.
  * @returns
  */
-function Map({ onLocationClick, onVenueChangedOnMap, useMapProviderModule }) {
+function Map({ onLocationClick, onVenueChangedOnMap, useMapProviderModule, onMapPositionKnown }) {
```
andreeaceachir142, line 39: Add function docs for the `onMapPositionKnown` prop 😄

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -24,6 +24,7 @@ import defaultLogo from "../../assets/logo.svg";
  * @param {string} [props.gmMapId] - The Google Maps Map ID associated with a specific map style or feature.
  * @param {boolean} [props.useMapProviderModule] - Set to true if the Map Template should take MapsIndoors solution modules into consideration when determining what map type to use.
  * @param {string} [props.kioskOriginLocationId] - If running the Map Template as a kiosk (upcoming feature), provide the Location ID that represents the location of the kiosk.
+ * @param {number} [props.timeout] - If you want the Map Template to reset map position and UI elements to the initial state after some time of inactivity, use this to specify the number of seconds of inactivity before resetting.
```
andreeaceachir142, line 27: Does the timeout prop only work in a kiosk context? If yes, it should probably be specified here

packages/map-template/src/helpers/GetDesktopPaddingBottom.js
```diff
@@ -1,9 +1,27 @@
...
+        if (!sidebar) {
+            const observer = new MutationObserver(() => {
+                const sidebar = document.querySelector('.modal--open');
+                if (sidebar) {
+                    observer.disconnect();
+                    resolve(mapContainer.offsetHeight - sidebar?.offsetTop); // Subtract the top padding from the height of the map container element.
+                }
+            });
+
+            observer.observe(document, { attributes: false, childList: true, characterData: false, subtree: true });
```
andreeaceachir142, line 22: Can you put a few more words to why this is needed? 🙏🏻

packages/map-template/src/hooks/useInactive.js
```diff
@@ -0,0 +1,59 @@
...
+import { useRecoilValue } from 'recoil';
+import isMapReadyState from '../atoms/isMapReadyState';
+import timeoutState from '../atoms/timoutState';
+
+/**
+ * Detect user inactivity and sets state to be inactive after a specified amount of seconds.
+ *
+ * Heavily inspired by useIdle from https://www.npmjs.com/package/@uidotdev/usehooks
+ *
+ * @param {number} seconds
```
andreeaceachir142, line 11: Is this still needed?

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -0,0 +1,289 @@
...
+     * When relevant state changes, run code to go to a location in the world.
+     */
+    useEffect(() =>  {
+        determineMapBounds();
+    }, [mapsIndoorsInstance, currentVenueName, venues, locationId, kioskOriginLocationId, pitch, bearing, startZoomLevel]);
+
+
+    /**
+     * Based on the combination of the states for venueName, locationId & kioskOriginLocationId,
+     * determine what to make the map go to.
```
andreeaceachir142, line 75:
```suggestion
     * determine where to make the map go to.
```

packages/map-template/src/helpers/GetDesktopPaddingBottom.js
ammapspeople: I tried in line 12. Or is it this specific line you want some elaboration on?

packages/map-template/src/components/Map/Map.jsx
ammapspeople: 🫡

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
ammapspeople: No, it doesn't require a kiosk context in accordance with the spec.

packages/map-template/src/hooks/useInactive.js
ammapspeople: No 👀

packages/map-template/src/helpers/GetDesktopPaddingBottom.js
andreeaceachir142: Yes, I was mostly asking about the props inside the object, childList, characterData, subtree

ammapspeople: I'll see what I can do. If anything, I can link to https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#options

andreeaceachir142: That's what I was looking for, perfect 😄

andreeaceachir142 (approved):
It looks good to me 👏🏻 Great work!

packages/map-template/src/helpers/GetDesktopPaddingBottom.js
ammapspeople: Done.

ammapspeople:
Part of another branch now. Closing.

PR #304, opened 2024-01-02
Add support for loading wayfiding when having directionsFrom prop

andreeaceachir142:
# What 

- Add support for loading the Wayfinding view when having the `directionsFrom` prop

# How

- Add an extra condition in the Sidebar and Bottom sheet components which checks if the `directionsFrom` parameter exists, and then load the Wayfinding with the prefilled value


ammapspeople (approved):
👏

PR #308, opened 2024-01-10
Fix logo and host it on S3 bucket 

andreeaceachir142:
# What 

- Host logo on S3 bucket due to limitation cause by having it under the assets folder


ammapspeople (approved):
👏 💨

PR #310, opened 2024-01-15
Add horizontal scroll buttons to categories list

andreeaceachir142:
# What 

- Add horizontal scroll buttons to categories list 
- Restrict the buttons to be shown only on desktop mode


packages/map-template/src/components/Search/Search.jsx
```diff
@@ -322,20 +368,38 @@ function Search({ onSetSize }) {
...
-                                background-color={primaryColor}
-                                content={categoryInfo.displayName}
-                                active={selectedCategory === category}
-                                onClick={() => categoryClicked(category)}
-                                key={category}>
-                            </mi-chip>
-                        )}
-                    </div>}
+            <div className="prevent-scroll" {...scrollableContentSwipePrevent}>
+                <div className="search__scrollable">
```
ammapspeople, line 372: Can we extract all JSX and logic related to the categories into a separate component? I think that would be a clear separation of concerns.

andreeaceachir142: Just the categories or the categories with scroll buttons as well?

andreeaceachir142: Great idea, refactored the code and extracted the categories into its own component. Now the search component has been simplified a lot.

packages/map-template/src/components/Categories/Categories.jsx
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Show the categories list.
+ * 
+ * @param {object} props
+ * @param {function} props.onSetSize - Callback that is fired when the categories are clicked.
+ * @param {function} props.getFilteredLocations - Function that gets the filtered locations based on the category selected.
+ * @param {object} props.searchFieldRef - The reference to the search input field.
+ */
+function Categories({ onSetSize, getFilteredLocations, searchFieldRef }) {
```
ammapspeople, line 22: About the placement of this component file, would it make sense to place it inside the Search component folder? That way we signal that this component is only used by the Search component. We do the same for the Sidebar & Modal components.

andreeaceachir142: Yes, that makes sense. 🙏🏻 Moved it to the Search folder.

PR #311, opened 2024-01-15
Use correct language notation for Mapbox search suggestions

ammapspeople:
When getting Mapbox search suggestions, reduce the set language string to the primary language subtag, which is seemingly the only way Mapbox will accept the language parameter.


andreeaceachir142 (approved):
Approved 👏🏻

andreeaceachir142:
Remember the Changelog and label before merging into main 🙏🏻 @ammapspeople

PR #314, opened 2024-01-22
Reset origin location when exiting wayfinding

andreeaceachir142:
# What 

- Reset origin location when exiting Wayfinding page 

# How 

- Create function that handles the closing of the Wayfinding page 
- Reset the origin location and also the display text 
- Adjust the `setDisplayText()` method on the `search.jsx` component and set the `inputElement.value` to also be equal to the `displayText`, otherwise the input field would keep the old value


ammapspeople (approved):
👏

PR #315, opened 2024-01-22
Scroll buttons in search results in kiosk mode

ammapspeople:
**Draft: Points towards feature branch**

# What

- Add scroll buttons when in kiosk mode to allow to scroll in the list of search results by clicking
- Change the scroll-able area in the search page to be search results only

# How

- Modified the Search component to change which area is scroll-able, and added more code whitespace in the JSX to (hopefully) make it easier to get an overview of.
- Use the `mi-scroll-buttons` component with a React portal to show the scroll buttons
- Add ResizeObserver to the scroll buttons component to update scroll buttons state whenever the dimensions of the scroll container changes
- Define CSS Parts on the scroll buttons component to allow for easy styling from the outside.


packages/map-template/src/components/Search/Search.scss
```diff
@@ -28,11 +25,61 @@
...
+            content: url('../../assets/arrow-down.svg');
+        }
+
+        mi-scroll-buttons::part(button-up):disabled::after {
+            content: url('../../assets/arrow-up-disabled.svg');
+        }
+
+        mi-scroll-buttons::part(button-down):disabled::after {
+            content: url('../../assets/arrow-down-disabled.svg');
+        }
```
andreeaceachir142, line 83: What do we need this for? 🤔

andreeaceachir142:
When clicking on the scroll buttons and having the keyboard open, I think the keyboard should still be visible 
Screen Recording 2024-01-22 at 14.45.34

packages/map-template/src/components/Search/Search.scss
ammapspeople: It's overriding the default looks of the scroll buttons component.

andreeaceachir142: Wouldn't it be a good idea to update the component instead? I will have to use the same buttons for the legend, and we'll end up with repetitive code

ammapspeople: I was reluctant to break the current layout of the component just for this one case, but since we will be using them in another place, I am ready to make the leap 😄 I'll update the component.

ammapspeople: 👍 Did some refactor.

ammapspeople:
> When clicking on the scroll buttons and having the keyboard open, I think the keyboard should still be visible

We decided to leave it as it is now: clicking scroll buttons (thus making the input field no longer have focus) removes the keyboard.

andreeaceachir142 (approved):
Approved ✅

PR #318, opened 2024-01-23
Fixed the directions not showing on mobile

andreeaceachir142:
# What 

- Fixed directions not being shown on mobile 

# How 

- Add missing code from resolving the promise when getting mobile bottom padding


ammapspeople (approved):
Nice catch!

PR #319, opened 2024-01-23
Fix: Pan the map to correct location when clicking on search result not belonging to the current venue

ammapspeople:
# What

- Fix bug where (when using `kioskOriginLocationId`) clicking on a search result not belonging to the currently set venue would not pan the map to the clicked Location, but instead reset the map to the initial settings.

# How

- Made the `useMapBoundsDeterminer` hook react to changes in the "venue" prop instead of the internal `venueName` state. Also manually trigger map pan when using the Venue Selector, since the responsibility of doing that was previously mistakenly given to the `useMapBoundsDeterminer` hook.

### DIFF
--- a/packages/map-template/src/hooks/useMapBoundsDeterminer.js
+++ b/packages/map-template/src/hooks/useMapBoundsDeterminer.js
@@ -70,7 +70,7 @@ const useMapBoundsDeterminer = () => {
      */
     useEffect(() =>  {
         determineMapBounds();
-    }, [mapsIndoorsInstance, currentVenueName, venues, locationId, kioskOriginLocationId, pitch, bearing, startZoomLevel, categories]);
+    }, [mapsIndoorsInstance, venue, venues, locationId, kioskOriginLocationId, pitch, bearing, startZoomLevel, categories]);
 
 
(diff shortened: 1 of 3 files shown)


packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -70,7 +70,7 @@ const useMapBoundsDeterminer = () => {
      */
     useEffect(() =>  {
         determineMapBounds();
-    }, [mapsIndoorsInstance, currentVenueName, venues, locationId, kioskOriginLocationId, pitch, bearing, startZoomLevel, categories]);
```
andreeaceachir142, line 73: We don't need the current venue here anymore?

ammapspeople: No, we only need the venue (the value of the venue prop that is externally exposed).

andreeaceachir142: Alright 😄

andreeaceachir142 (approved):
👏🏻

PR #320, opened 2024-01-24
Horizontal and vertical scroll bars on the Kiosk

andreeaceachir142:
# What 

- This PR is a combination of both the vertical and horizontal scroll bars


packages/components/src/components/scroll-buttons/test.html
```diff
@@ -18,7 +18,7 @@
 </head>
 
 <body>
-    <section style="display: flex;">
+    <section style="display: flex; background-color: #eee;">
```
andreeaceachir142, line 21: Do we need this change here? 🤔

ammapspeople: For internal testing. The buttons are now white, which makes them hard to see on a white background.

andreeaceachir142:
@ammapspeople Added new icons, removed the scroll bar and aligned the buttons with the categories. Ready for re review 🙏🏻

packages/map-template/src/components/Search/Categories/Categories.scss
```diff
@@ -3,16 +3,21 @@
     flex-shrink: 0;
     display: grid;
     grid-auto-flow: column;
-    align-items: end;
     gap: var(--spacing-x-small);
+    margin-top: var(--spacing-x-small);
 
+    ::-webkit-scrollbar {
```
ammapspeople, line 9: Will this work in Firefox and Safari?

ammapspeople: I may just answer this myself: It does not work in Firefox 😄

andreeaceachir142: Can you send me a screenshot of how it looks on Firefox please? 🙏🏻

ammapspeople: No, because you fixed it 😄

ammapspeople (commented):
👍

PR #323, opened 2024-02-01
Remove call for resetting the tilt when having an untracked position

andreeaceachir142:
# What 

- Fix issue of map resetting the tilt when having the position enabled 

# How

- Remove call for resetting the tilt when having the position enabled but untracked


ammapspeople (approved):
👏

PR #324, opened 2024-02-02
Add support for searching within the same venue in kiosk mode

andreeaceachir142:
# What 

- Add support for searching for locations which are in the same venue in the kiosk mode

# How

- Add `mi-venue` attribute to the `mi-search` component which checks if kiosk we are in kiosk mode
- Add `venue` attribute when getting the filtered locations after clicking on a location 
- Add the `venueId` from the kiosk location to both instances

### DIFF
--- a/packages/map-template/src/components/Search/Search.jsx
+++ b/packages/map-template/src/components/Search/Search.jsx
@@ -325,7 +325,7 @@ function Search({ onSetSize, isOpen }) {
 
 
             { /* Search field that allows users to search for locations (MapsIndoors Locations and external) */}
-            
+
             <SearchField
                 ref={searchFieldRef}
                 mapsindoors={true}
@@ -351,11 +351,11 @@ function Search({ onSetSize, isOpen }) {
             { /* Message shown if no search results were found */}
 
             {showNotFoundMessage && <p className="search__error"> {t('Nothing was found')}</p>}
-            
-            
+
+
 
             { /* Vertical list of search results. Scrollable. */}
-            
+
             {searchResults.length > 0 &&
                 <div className="search__results prevent-scroll" {...scrollableContentSwipePrevent}>
                     {searchResults.map(location =>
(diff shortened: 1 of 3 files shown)


packages/map-template/src/components/Search/Search.jsx
```diff
@@ -316,75 +317,49 @@ function Search({ onSetSize, isOpen }) {
...
     }, [useKeyboard]);
 
-
     return (
         <div className="search"
             ref={searchRef}
-            style={calculateContainerStyle()}
-            >
-
-                { /* Search field that allows users to search for locations (MapsIndoors Locations and external) */ }
```
ammapspeople, line 326: Aww, I liked the whitespace that made it easier to read 😢 

Did you change anything else but removing empty lines?

andreeaceachir142: I knoow... I liked the readability as well, but it's inconsistent with everything else and also a bit hard to maintain when the formatter goes against me 🙃

andreeaceachir142: Nothing else was changed, except the empty lines.

ammapspeople: Don't let the editor win against you 😄
It's not inconsistent with everything else: look at LocationDetails.jsx where we have a similar-looking pattern for readability. Also, we need to make the code easier to read.

packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
```diff
@@ -112,6 +118,7 @@ const SearchField = forwardRef(({ placeholder, mapsindoors, results, clicked, cl
         disabled={disabled}
         mapbox={mapbox}
         google={google}
+        mi-venue={kioskLocation && isKioskContext ? kioskLocation.properties.venueId : undefined}
```
ammapspeople, line 121: Add a comment explaining that when the Map Template runs in kiosk mode, we restrict searching to the venue only.

packages/map-template/src/components/Search/Search.jsx
ammapspeople: If you just leave a little white space for readability, I'm happy 😄

ammapspeople (approved):
👏

PR #325, opened 2024-02-06
Add legend modal to the kiosk

andreeaceachir142:
# What

- Add legend modal to the kiosk

# How

- Create a `LegendDialog.jsx` component which handles the legend content
- Use Mutation Observer to get the Legend sections height
- Create helper method which returns the legend sorted fields


ammapspeople:
I see some weird scrollbar/shadows when the legend is short

image

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -316,14 +324,30 @@ function Search({ onSetSize, isOpen }) {
         }
     }, [useKeyboard]);
 
+    /*
+     * Get the legend sections and determine 
+     * If the legend button should be shown.
+     */
+    useEffect(() => {
+        const sections = getLegendSections(kioskLocation);
+        if (sections.length > 0) {
```
ammapspeople, line 333: This crashes the map template in case there are no Legend sections.

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const [showScrollButtons, setShowScrollButtons] = useState(false);
+
+    /*
+     * Get the height of the legend sections. 
+     * Set the legend size atom and determine 
+     * if the scroll buttons should be shown.
+     */
+    useEffect(() => {
+        if (showLegendDialog && isKioskContext) {
+            getLegendSectionsHeight().then(padding => {
```
ammapspeople, line 47: "Height"? "Padding"? "Size"? Choose one 😄

packages/map-template/src/helpers/GetLegendSectionsHeight.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Get height of the legend sections.
+ * Returns a promise with the height in pixels.
+ *
+ * @returns {Promise}
+ */
+export default function getLegendSectionsHeight() {
```
ammapspeople, line 7: This is related to the LegendDialog component only. I think this file should be moved into the LegendDialog folder.

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -0,0 +1,93 @@
...
+                    {legendSections.map((legendSection, index) => <div key={index} className="legend__section" ref={legendSectionRef}>
+                        <div className="legend__heading">{legendSection.heading}</div>
+                        <div className="legend__content">{legendSection.content}</div>
+                    </div>)}
+                </div>}
+            <button className="legend__button" style={{ background: primaryColorProp }} onClick={() => setShowLegendDialog(false)}>{t('Close')}</button>
+        </div>
+
+        { /* Buttons to scroll in the list of search results if in kiosk context */}
+        {isKioskContext && showLegendDialog && showScrollButtons && createPortal(
```
ammapspeople, line 83: Is the createPortal necessary when it's placed within the same React component?

And if it can be removed, will that solve the issue that we introduced the MutationObserver for?

packages/map-template/src/components/LegendDialog/LegendDialog.scss
```diff
@@ -0,0 +1,76 @@
+.background {
+    z-index: 10000;
```
ammapspeople, line 2: Can we use one of our existing variables for z-index?

packages/map-template/src/components/LegendDialog/LegendDialog.scss
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    &__sections {
+        overflow: scroll;
+
+        &--scrollable {
+            &::before {
+                content: '';
+                position: absolute;
+                bottom: calc(64px + var(--spacing-small));
```
ammapspeople, line 62: Should there be a similar gradient for the top?

packages/map-template/src/helpers/GetLegendSortedFields.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Get the legend sorted fields.
+ *
+ * @param {object} kioskLocation
+ * @returns {Promise}
+ */
+export default function getLegendSortedFields(kioskLocation) {
```
ammapspeople, line 7: I'm wondering if this could be a Recoil Selector that depends on the kioskLocation atom.

packages/map-template/src/helpers/GetLegendSectionsHeight.js
andreeaceachir142: ✅

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: Added check for existence of kiosk location

packages/map-template/src/components/LegendDialog/LegendDialog.scss
andreeaceachir142: ✅

packages/map-template/src/components/LegendDialog/LegendDialog.scss
andreeaceachir142: Hmmm, not according to the Figma design 🤔

packages/map-template/src/helpers/GetLegendSortedFields.js
andreeaceachir142: Hmm, I can't see how that would really work. We use the kioskLocation in many places, and the legend is only used in two places.

ammapspeople: I don't see how converting getLegendSortedFields to a Recoil Selector (which would also use the kioskLocation atom) could affect the use of the kioskLocation atom. Can you elaborate?

packages/map-template/src/components/LegendDialog/LegendDialog.scss
ammapspeople: Weird. We have it for long location detail descriptions 🤷

andreeaceachir142: I'll check with UX 👍🏻

packages/map-template/src/atoms/showLegendDialogState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const showLegendDialogState = atom({
```
ammapspeople, line 3: Such booleans I prefer to prefix with "is" (like isMapReadyState, isLocationClickedState etc.).
Can we name this following that pattern? Something like isLegendDialogVisible?

andreeaceachir142: Good idea ✅

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
andreeaceachir142: I think the createPortal and the Mutation observer are not necessary anymore. Refactored the code and pushed the changes.

andreeaceachir142:
@ammapspeople Can you still see the scroll bars?

packages/map-template/src/helpers/GetLegendSortedFields.js
andreeaceachir142: Used Recoil Selector ✅

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
ammapspeople: 🥳

ammapspeople:
> @ammapspeople Can you still see the scroll bars?

No, not anymore 😄 

I do however see that the scrollbar is also covered by the shadow effect. Is that intended?
image

packages/map-template/src/atoms/showLegendDialogState.js
ammapspeople: 🎉

packages/map-template/src/components/LegendDialog/LegendDialog.scss
ammapspeople: Did you get an answer?

andreeaceachir142: Yess, it should also be on top, so I will implement that 😄

andreeaceachir142: Added gradient on the top of the legend dialog, when scrollable.

packages/map-template/src/atoms/legendHeightState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const legendHeightState = atom({
```
ammapspeople, line 3: We use it in the LegendDialog component only. Consider using a useState hook instead.

andreeaceachir142: ✅

andreeaceachir142:
Added the correct oder of z-index on the different components and restrict the users to interact with the rest of the map when having the legend modal open. @ammapspeople

ammapspeople:
Seems like the dimmed scrollbar is also an issue for location details, and that has never been reported by anyone 🙃 

However, I just noticed another problem: The legend is not closed by the timeout.

andreeaceachir142:
Oh 👀 Looking into that now. I assume the same issue happens when we have the QR code dialog open too @ammapspeople I'll fix both of them.

andreeaceachir142:
Fixed ✅ @ammapspeople

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+    /*
+     * Get the height of the legend sections. 
+     * Set the legend size atom and determine 
+     * if the scroll buttons should be shown.
+     */
+    useEffect(() => {
+        if (showLegendDialog && isKioskContext) {
+            if (legendSectionRef.current.clientHeight > 700) {
+                setLegendHeight(legendSectionRef.current.clientHeight);
```
ammapspeople, line 44: Why is it that we only set this if the height is larger than 700?
I see where we use the value, that we check if it is larger than 700. Should the value of the useState be a boolean instead of the height?

packages/map-template/src/components/LegendDialog/LegendDialog.scss
```diff
@@ -0,0 +1,83 @@
+.background {
```
ammapspeople, line 1: Should we prefix it with something like legend to avoid it being "global"? Can we use BEM?

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -317,25 +327,43 @@ function Search({ onSetSize, isOpen }) {
...
+     * Get the legend sections and determine 
+     * If the legend button should be shown.
+     */
+    useEffect(() => {
+        if (kioskLocation) {
+            if (legendSections.length > 0) {
+                setShowLegendButton(true);
+            } else {
+                setShowLegendButton(false);
+            }
```
ammapspeople, line 340:
```suggestion
            setShowLegendButton(legendSections.length > 0);
```

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -348,6 +376,8 @@ function Search({ onSetSize, isOpen }) {
 
 
 
+            { /* Horizontal list of Categories */}
```
ammapspeople, line 379: Seems like this doesn't belong here 😄

packages/map-template/src/components/Search/Search.scss
```diff
@@ -18,7 +18,7 @@
         flex: 1 1 auto;
         margin-top: var(--spacing-x-small);
 
-        > * {
+        >* {
```
ammapspeople, line 21:
```suggestion
        > * {
```

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
andreeaceachir142: We use the 700 here because that is the UX spec. 
Hmm, how do you imagine it to be a boolean? How would we check for the height?

ammapspeople: `setLegendIsTall(legendSectionRef.current.clientHeight > 700)`

(find a better name though)

Then we also avoid having the number 700 hardcoded in more than one place.

packages/map-template/src/components/LegendDialog/LegendDialog.scss
andreeaceachir142: ✅

packages/map-template/src/components/Search/Search.jsx
andreeaceachir142: 🫣 Removed it

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
andreeaceachir142: Ah, great idea! I'll refactor that

andreeaceachir142: ✅

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -0,0 +1,88 @@
...
+     * if the scroll buttons should be shown.
+     */
+    useEffect(() => {
+        if (showLegendDialog && isKioskContext) {
+            if (legendSectionRef.current.clientHeight > 700) {
+                setIsLegendScrollable(true);
+                setShowScrollButtons(true);
+            } else {
+                setShowScrollButtons(false);
+            }
```
ammapspeople, line 48: Can the `isLegendScrollable` and `showScrollButtons` be merged into one? They seem to have the same values.

andreeaceachir142: That's true, great idea! ✅

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -0,0 +1,81 @@
...
+    const legendSectionsRef = useRef();
+
+    /*
+     * Get the height of the legend sections. 
+     * Set the legend size atom and determine 
+     * if the scroll buttons should be shown.
+     */
+    useEffect(() => {
+        if (showLegendDialog && isKioskContext) {
+            legendSectionRef.current.clientHeight > 700 ? setIsLegendScrollable(true) : setIsLegendScrollable(false);
```
ammapspeople, line 41:
```suggestion
            setIsLegendScrollable(legendSectionRef.current.clientHeight > 700);
```

packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }, [showLegendDialog, legendSections, isLegendScrollable]);
+
+    return (<>
+        <div className="legend__background"></div>
+        <div className="legend" ref={legendModalRef}>
+            {legendSections.length > 0 &&
+                <div className={`legend__sections ${isLegendScrollable ? 'legend__sections--scrollable' : ''}`}
+                    ref={legendSectionsRef}
+                    style={{ maxHeight: isLegendScrollable ? '700px' : 'auto', overflowY: isLegendScrollable ? 'scroll' : 'auto' }}>
```
ammapspeople, line 61: The value of 700 is used in more than one case. Consider using a variable.

andreeaceachir142: Good idea, it's used in two cases, once just `700` and once `700px`. Does it make sense to use a variable in this case?

ammapspeople: Sure. 

We can always use template literals to get the px string: `${myAwesomeVariable}px`.

ammapspeople: 🙌

ammapspeople (approved):
👏

PR #326, opened 2024-02-06
Hide zoom controls from mobile view

andreeaceachir142:
# What 

- Hide zoom controls from mobile view

# How 

- Use the `isDesktop` media query to check if the zoom control should be displayed or not

### DIFF
--- a/packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
+++ b/packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
@@ -8,6 +8,7 @@ import primaryColorState from '../../../atoms/primaryColorState';
 import gmMapIdState from '../../../atoms/gmMapIdState';
 import bearingState from '../../../atoms/bearingState';
 import pitchState from '../../../atoms/pitchState';
+import useMediaQuery from '../../../hooks/useMediaQuery';
 
 /**
  * Takes care of instantiating a MapsIndoors Google Maps MapView.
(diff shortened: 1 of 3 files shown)


packages/map-template/src/components/Map/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -29,6 +30,7 @@ function GoogleMapsMap({ onMapView, onPositionControl }) {
     const primaryColor = useRecoilValue(primaryColorState);
     const bearing = useRecoilValue(bearingState);
     const pitch = useRecoilValue(pitchState);
+    const isDesktop = useMediaQuery('(min-width: 992px)');
```
ammapspeople, line 33: We now have this `min-width: 992px` in many places, making it harder to change. Can we do anything about that?

andreeaceachir142: Maybe we could have a helper function instead which would return a boolean if it's desktop? What do you think?

ammapspeople: Something like that yes 😄

ammapspeople: Or a new custom hook that uses the useMediaQuery hook.

andreeaceachir142: Refactored the code and added a custom hook that uses the useMediaQuery hook.

ammapspeople: 👏 🙌

ammapspeople (approved):
✅

PR #327, opened 2024-02-06
Fix error when scanning the QR code

andreeaceachir142:
# What 

- Fix the issue of having incomplete URL when scanning the QR code

# How 

- Combine the `origin` and `pathname` when creating the targetURL


ammapspeople (approved):
👍

PR #328, opened 2024-02-06
Search results state is not reset when using the timeout property

andreeaceachir142:
# What 

- Fixed search results not being reset when using the timeout property 

# How

- Add the `searchResults` to the `useReset` hook


ammapspeople (approved):
🙌

PR #329, opened 2024-02-06
Accessibility not resetting when using the timeout property

andreeaceachir142:
# What 

- Fix accessibility not resetting when using the timeout property 

# How 

- Add the `accessibilityOn` state on the `useReset` hook


ammapspeople (approved):
👌

PR #330, opened 2024-02-07
Check if the pathname is a / character.

andreeaceachir142:
# What 

- Fix QR code error


packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -34,7 +34,9 @@ function QRCodeDialog() {
         if (directionsFrom && directionsTo) {
             // The target URL when the user opens the QR code dialog
             // Create the URL by combining the origin and the pathname
-            let targetUrl = window.location.origin + window.location.pathname;
+            // Check if the pathname is a '/' before appending it.
+
```
ammapspeople, line 38:
```suggestion
```

ammapspeople (approved):
👏

PR #331, opened 2024-02-07
Check for last character of the path name

andreeaceachir142:
# What 

- Fix error when having extra `/` character

# How 

- Check if the last character of the pathname is a `/` and remove it

### DIFF
--- a/packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
+++ b/packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
@@ -33,13 +33,9 @@ function QRCodeDialog() {
     useEffect(() => {
         if (directionsFrom && directionsTo) {
             // Get the last character of the pathname
-            const lastCharacter = window.location.pathname.slice(-1);
-
-            // Check if the pathname is a '/' before appending it to the targetUrl
-            let pathname = (lastCharacter === '/') ? window.location.pathname.replace(/\/$/, "") : window.location.pathname;
-             
             // Create the target URL when the user opens the QR code dialog
-            let targetUrl = window.location.origin + pathname;
+            // Replace the '/' at the end of the line
+            let targetUrl = window.location.origin + window.location.pathname.replace(/\/$/, "");
             
             // The interface for the existing URL search params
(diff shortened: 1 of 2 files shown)


packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -32,11 +32,15 @@ function QRCodeDialog() {
...
         if (directionsFrom && directionsTo) {
-            // The target URL when the user opens the QR code dialog
-            // Create the URL by combining the origin and the pathname
-            // Check if the pathname is a '/' before appending it.
-            let targetUrl = window.location.origin + (window.location.pathname === '/' ? '' : window.location.pathname)
+            // Get the last character of the pathname
+            const lastCharacter = window.location.pathname.slice(-1);
 
+            // Check if the pathname is a '/' before appending it to the targetUrl
+            let pathname = (lastCharacter === '/') ? window.location.pathname.slice(0, -1) : window.location.pathname;
```
ammapspeople, line 39: Going with a `replace` using a regular expression may be more elegant: https://stackoverflow.com/a/7602314

Either way, this needs some good testing.

andreeaceachir142: That's a good idea, I used the `replace` instead.

packages/map-template/src/components/QRCodeDialog/QRCodeDialog.jsx
```diff
@@ -32,11 +32,15 @@ function QRCodeDialog() {
 
     useEffect(() => {
         if (directionsFrom && directionsTo) {
-            // The target URL when the user opens the QR code dialog
-            // Create the URL by combining the origin and the pathname
-            // Check if the pathname is a '/' before appending it.
-            let targetUrl = window.location.origin + (window.location.pathname === '/' ? '' : window.location.pathname)
+            // Get the last character of the pathname
+            const lastCharacter = window.location.pathname.slice(-1);
```
ammapspeople, line 36: I don't think we need this anymore. The regular expression should take care of it.

ammapspeople (approved):
👏

PR #332, opened 2024-02-09
Add support for category property

andreeaceachir142:
# What 

- Add support for category property 

# How 

- Add new property called `category`
- Check if the category prop exists in the list of sorted categories -> if it exists, then set the `selectedCategory` atom 
- Use `MutationObserver` to detect when the active chip element appears on the map 
- Scroll into the view of the active category 

**Note!** Remember that the category property should be the Key (Administrative ID) and not the Display name.

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -76,7 +76,7 @@ defineCustomElements();
  * @param {boolean} [props.supportsUrlParameters] - Set to true if you want to support URL Parameters to configure the Map Template.
  * @param {boolean} [props.useKeyboard] - If running the Map Template as a kiosk, set this prop to true and it will prompt a keyboard.
  * @param {number} [props.timeout] - If you want the Map Template to reset map position and UI elements to the initial state after some time of inactivity, use this to specify the number of seconds of inactivity before resetting.
- * @param {string} [props.category] - If you want to indicate an active category on the map.
+ * @param {string} [props.category] - If you want to indicate an active category on the map. The value should be the Key (Administrative ID).
  */
 function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, primaryColor, logo, appUserRoles, directionsFrom, directionsTo, externalIDs, tileStyle, startZoomLevel, bearing, pitch, gmMapId, useMapProviderModule, kioskOriginLocationId, language, supportsUrlParameters, useKeyboard, timeout, category }) {
 
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -563,13 +563,8 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
      * Check if the category property matches with any of the existing categories.
      */
     useEffect(() => {
-        if (mapsindoorsSDKAvailable) {
-            if (category) {
-                const existingCategory = categories.find((matched) => matched[0] === category);
-                if (existingCategory !== undefined) {
-                    setSelectedCategory(category);
-                }
-            }
+        if (mapsindoorsSDKAvailable && category && categories.find((matched) => matched[0] === category)) {
+            setSelectedCategory(category);
         }
     }, [category, categories, mapsindoorsSDKAvailable]);
 
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -46,6 +46,7 @@ import venueState from '../../atoms/venueState.js';
 import useSetCurrentVenueName from '../../hooks/useSetCurrentVenueName.js';
 import useKeyboardState from '../../atoms/useKeyboardState';
 import { useIsDesktop } from '../../hooks/useIsDesktop.js';
+import selectedCategoryState from '../../atoms/selectedCategoryState.js';
 
 // Define the Custom Elements from our components package.
 defineCustomElements();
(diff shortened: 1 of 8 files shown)


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -544,6 +547,21 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
     }
 
+    /*
+     * React on changes in the category prop.
+     * Check if the category property matches with any of the existing categories.
+     */
+    useEffect(() => {
+        if (mapsindoorsSDKAvailable) {
+            if (category) {
+                const existingCategory = categories.find((matched) => matched[0] === category)
```
ammapspeople, line 557: Can probably be shortened (not tested)

```suggestion
            if (category && categories.includes(matched => matched[0] === category) {
```

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -124,6 +126,25 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef }) {
         });
     }
 
+    /**
+     * Get the active category element.
+     */
+    useEffect(() => {
+        if (selectedCategory) {
+            getActiveCategory().then(activeCategory => setActiveCategory(activeCategory));
```
ammapspeople, line 134: Can we listen for the categories atom instead of using a mutationobserver?

andreeaceachir142: I'm not sure, because we need to determine which one is the active category, and that's why we're using the Mutation Observer here.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -544,6 +547,20 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setCategories(uniqueCategories);
     }
 
+    /*
+     * React on changes in the category prop.
+     * Check if the category property matches with any of the existing categories.
+     */
+    useEffect(() => {
+        if (mapsindoorsSDKAvailable) {
+            if (category && categories.includes(matched => matched[0] === category) {
```
ammapspeople, line 556: The code cannot compile due to a syntax error here. When I say my suggestions are not tested, I really mean it 😅

packages/map-template/src/components/Search/Categories/Categories.jsx
ammapspeople: Try playing around with making a conditional ref (or whatever one could call it) for the active chip.

In the JSX where we loop through the categories and present `<mi-chip>` elements, add something like

`const refProp = selectedCategory === category ? { ref: activeChipRef } : {};`

And add `{...refProp}` as an attribute to the `<mi-chip>` element.

Then it may (this is not tested!! 👈 ) be possible to listen for the appearance of the `activeChipRef` in a `useEffect` hook and do the scrolling there. Without the use of a MutationObserver.

andreeaceachir142: I tried your suggestion, and it doesn't seem to work because the selected category is undefined at the time when we render the categories. 

I tried to fiddle around with it by creating another `useState` and assigning that at the moment when we have the `selectedCategory` present, but didn't get it to work. I might be doing something wrong.

I stashed my changes, we can take a look at it together whenever you have some time, and see if we can get it to work.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Reverted the changes 👍

packages/map-template/src/components/Search/Categories/Categories.jsx
ammapspeople: I just tried an actual implementation of my idea, and can't get it to work without introducing a `setTimeout` 😂 I think it may be solvable though (I am also a little confused about the difference between activeCategory and selectedCategory).

andreeaceachir142: Good old `setTimeout` 😂

andreeaceachir142: Should we see if we can get it to work though?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -556,6 +558,21 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+     */
+    useEffect(() => {
+        if (mapsindoorsSDKAvailable) {
+            if (category) {
+                const existingCategory = categories.find((matched) => matched[0] === category);
+                if (existingCategory !== undefined) {
+                    setSelectedCategory(category);
+                }
+            }
+        }
```
ammapspeople, line 573: Can it be shortened to (and please test 😅 )

```suggestion
        if (mapsindoorsSDKAvailable && category && categories.find((matched) => matched[0] === category)) {
            setSelectedCategory(category);
        }
```

packages/map-template/src/components/Search/Categories/Categories.jsx
ammapspeople: After spending some time on fixing this, we decided to postpone a fix, and use what we have now.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Yess, it seems to be working in the shortened version. ✅

andreeaceachir142: Pushed the changes from my local branch.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -124,6 +126,25 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef }) {
         });
     }
 
+    /**
```
ammapspeople, line 129:
```suggestion
    /*
```

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -124,6 +126,25 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef }) {
...
+     * Get the active category element.
+     */
+    useEffect(() => {
+        if (selectedCategory) {
+            getActiveCategory().then(activeCategory => setActiveCategory(activeCategory));
+        }
+    }, [selectedCategory]);
+
+    /**
```
ammapspeople, line 138:
```suggestion
    /*
```

ammapspeople (approved):
🥳

PR #337, opened 2024-02-16
'section' Prop for the data-table component

annakubackaAnku:
### What
Make data-table component work with pagination

### How
I created an addition property for this component called section which is used to display a certain section of the filtered items put into the table.

### DIFF
--- a/packages/components/src/components/data-table/data-table.tsx
+++ b/packages/components/src/components/data-table/data-table.tsx
@@ -58,6 +58,7 @@ export class DataTable {
      */
     @Watch('section')
     onSectionChangedHandler(): void {
+        this.currentSection = this.section;
         this.renderTableContent();
     }
 
@@ -87,6 +88,8 @@ export class DataTable {
     @Prop({ attribute: 'sticky-header' }) isHeaderSticky = true;
 
     @State() tableRows = [];
+
+    @State() currentSection = 1;
 
     /**
      * State to keep track of the table sorting.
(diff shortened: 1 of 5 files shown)


packages/components/src/components/data-table/test.html
```diff
@@ -34,6 +34,13 @@
         </mi-column>
     </mi-data-table>
 
+    <!-- <input type="number" id="myInput">
```
ammapspeople, line 37: This should be cleaned up.

ammapspeople: We could absolutely have the prop showcased on the test page, but it was commented out.

annakubackaAnku: I extended the test file then 😇  I just wanted to avoid changing the existing one, but now it's updated with showcasing the new property

packages/components/src/components/data-table/data-table.tsx
```diff
@@ -43,11 +43,25 @@ export class DataTable {
...
         this.setHeaderCheckboxState();
     }
+
     /**
      * The maximum number of rows to be displayed.
      */
     @Prop() maxRows: number;
 
+    /**
+     * The section of rows to be displayed.
```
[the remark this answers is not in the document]
ammapspeople: So page or section?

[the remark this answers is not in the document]
annakubackaAnku: renamed 👍

packages/components/src/components/data-table/test.html
ammapspeople: Should we showcase it the prop on the readme page too?

annakubackaAnku: I kind of forgot that existed :DD I added it now 👍

ammapspeople (approved):
👏 

Remember to set a label on the pull request and update the changelog before merging.

PR #339, opened 2024-02-22
Remove hover from non-selectable locations

andreeaceachir142:
# What 

- Remove hover from non-selectable locations 

# How 

- Implement functionality for checking if the location is non-selectable before calling the `hoverLocation()` and `unhoverLocation()` on the `ListItemLocation.jsx` file
- Implement functionality for detecting the hovered locations on the map, and if the location is non-selectable, call the `unhoverLocation()`

### DIFF
--- a/packages/map-template/src/components/Map/Map.jsx
+++ b/packages/map-template/src/components/Map/Map.jsx
@@ -203,7 +203,8 @@ function Map({ onLocationClick, onVenueChangedOnMap, useMapProviderModule, onMap
             mapView
         });
 
-        // If the location where the mouse entered is non-selectable, do not hover it.
+        // Detect when the mouse hovers over a location and store the hovered location
+        // If the location is non-selectable, remove the hovering by calling the unhoverLocation() method.
         miInstance.on('mouseenter', () => {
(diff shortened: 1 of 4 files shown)


packages/map-template/src/components/Map/Map.jsx
```diff
@@ -203,6 +203,16 @@ function Map({ onLocationClick, onVenueChangedOnMap, useMapProviderModule, onMap
             mapView
         });
 
+        // Detect when the mouse hovers over a location and store the hovered location
+        // If the location is non-selectable, remove the hovering by calling the unhoverLocation() method.
```
ammapspeople, line 207: So the SDK applies hover effects to non-selectable Locations out of the box, and we chose to make the Map Template prevent that behaviour.

Seems like an odd way of slicing the cake: Shouldn't this be ported to the SDK? 

CC @markhougaard

andreeaceachir142: This was exactly the question that came to my mind as well, when I started implementing this. I had a talk with @matbmapspeople who confirmed the instructions were to implement this on the Map Template instead of the SDK. 

cc @markhougaard @ammapspeople

PR #340, opened 2024-02-22
Fix size of the images on the Location Details page

andreeaceachir142:
# What 

- Fix size of the images on the Location Details prop

# How 

- Based on UX considerations, removed the `aspect-ratio`, changed the `object-fit` to `contain` and added a padding of `16px` at the top and bottom of the image.


andreeaceachir142:
Added new styles based on UX decision @ammapspeople

ammapspeople (approved):

PR #343, opened 2024-03-04
Add Sentry performance monitoring

markhougaard:
Adding support for Sentry Error Logging, Release Tracking, and Application Performance Monitoring.

One big thing that makes the Map Template deviate from the standard Sentry docs is that we don't use React Routes, but keep everything on `/` as a single-page app. We can do profiling of individual components, and that might give us more insight into which ones are heavy/slow, or other bottlenecks they might create. Docs here: https://docs.sentry.io/platforms/javascript/guides/react/features/component-tracking/. I haven't seen a named Component automatically tracked, besides the one [I made myself](https://github.com/MapsPeople/web-ui/pull/343#pullrequestreview-1914170685), so we need to go in and name all of the components.

### DIFF
--- a/packages/map-template/.gitignore
+++ b/packages/map-template/.gitignore
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# Sentry Config File
+.env.sentry-build-plugin
--- a/packages/map-template/src/App.jsx
+++ b/packages/map-template/src/App.jsx
@@ -14,5 +14,4 @@ function App() {
     );
 }
 
-export default Sentry.withProfiler(App, { name: "MapsIndoorsMap" });
-
+export default Sentry.withProfiler(App, { name: "App" });
--- a/packages/map-template/src/App.jsx
+++ b/packages/map-template/src/App.jsx
@@ -1,3 +1,4 @@
+import * as Sentry from "@sentry/react";
 import './App.css';
 import MapsIndoorsMap from './components/MapsIndoorsMap/MapsIndoorsMap';
 
@@ -13,5 +14,5 @@ function App() {
     );
 }
 
-export default App;
+export default Sentry.withProfiler(App, { name: "MapsIndoorsMap" });
 
--- a/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
+++ b/packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
@@ -1,4 +1,6 @@
-import { useEffect, useState } from 'react';
+import * as Sentry from "@sentry/react";
+import React, { useEffect, useState } from 'react';
+import { createRoutesFromChildren, matchRoutes, useLocation, useNavigationType } from 'react-dom/client';
 import { RecoilRoot } from 'recoil';
 import MapTemplate from '../MapTemplate/MapTemplate.jsx';
 
@@ -110,6 +112,36 @@ function MapsIndoorsMap(props) {
         </RecoilRoot>
     )
 }
+
+Sentry.init({
+    dsn: "https://0ee7fa162023d958c96db25e99c8ff6c@o351128.ingest.sentry.io/4506851619831808",
+    // Set environment to localhost if the url includes it. Otherwise, set to production.
+    environment: window.location.hostname === 'localhost' ? 'localhost' : 'production',
+    integrations: [
+        // See docs for support of different versions of variation of react router
+        // https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/
+        Sentry.reactRouterV6BrowserTracingIntegration({
+            useEffect: React.useEffect,
+            useLocation,
+            useNavigationType,
+            createRoutesFromChildren,
+            matchRoutes
+        }),
+        Sentry.replayIntegration()
+    ],
+
+    // Set tracesSampleRate to 1.0 to capture 100%
+    // of transactions for performance monitoring.
+    tracesSampleRate: 1.0,
+
+    // Set `tracePropagationTargets` to control for which URLs distributed tracing should be enabled
+    tracePropagationTargets: ["localhost", /^https:\/\/api\.mapsindoors\.com/],
+
+    // Capture Replay for 10% of all sessions,
+    // plus for 100% of sessions with an error
+    replaysSessionSampleRate: 0.1,
+    replaysOnErrorSampleRate: 1.0,
+});
 
 export default MapsIndoorsMap;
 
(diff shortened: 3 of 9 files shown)


packages/map-template/src/App.jsx
```diff
@@ -13,5 +14,5 @@ function App() {
     );
 }
 
-export default App;
+export default Sentry.withProfiler(App, { name: "MapsIndoorsMap" });
```
markhougaard, line 17: This is really neat; you can see exactly when a component has been loaded using the "Spans" feature in Sentry, and filtering by the `ui.react.mount` option here: https://mapspeople-gx.sentry.io/performance/summary/spans/ui.react.mount:3112e291a8e98fa9/?project=4506851619831808&query=&statsPeriod=1h&transaction=%2F

image

ammapspeople:
Does this work for the built Web component and React component? I see that a lot of configuration is done on the app level, but we only build and export the MapsIndoorsMap component. Is this configuration passed down?

packages/map-template/src/index.jsx
```diff
@@ -1,12 +1,43 @@
+import * as Sentry from "@sentry/react";
 import React from 'react';
-import ReactDOM from 'react-dom/client';
+import ReactDOM, { createRoutesFromChildren, matchRoutes, useLocation, useNavigationType } from 'react-dom/client';
 import App from './App';
 import './index.css';
 import reportWebVitals from './reportWebVitals';
 
+Sentry.init({
+    dsn: "https://0ee7fa162023d958c96db25e99c8ff6c@o351128.ingest.sentry.io/4506851619831808",
```
ammapspeople, line 9: Is this a secret that is now no longer a secret? 🙃 Should it be configurable for customers, or is it something that we own and manage 100%?

packages/map-template/src/index.jsx
```diff
@@ -1,12 +1,43 @@
...
+    // Set environment to localhost if the url includes it. Otherwise, set to production.
+    environment: window.location.hostname === 'localhost' ? 'localhost' : 'production',
```
ammapspeople, line 11: This will set it to production when we test on mobile devices using Vite's `--host` flag. In that case, we access an  IP, and not `localhost`. Can we live with that?

packages/map-template/src/index.jsx
```diff
@@ -1,12 +1,43 @@
...
+    integrations: [
+        // See docs for support of different versions of variation of react router
```
ammapspeople, line 13: I'm 100% sure there is a good explanation for including some React Router related stuff here, but I would very much like to understand it, since we don't use a router. Can you elaborate in the comment?

packages/map-template/src/index.jsx
markhougaard: @ammapspeople It's not a secret. Sentry writes [in their docs](https://docs.sentry.io/product/sentry-basics/concepts/dsn-explainer/#dsn-utilization):

> DSNs are safe to keep public because they only allow submission of new events and related event data; they do not allow read access to any information.
>
> While there is a risk of abusing a DSN, where any user can send events to your organization with any information they want, this is a rare occurrence. Sentry provides controls to [block IPs](https://docs.sentry.io/platform-redirect/?next=/configuration/options/) and similar concerns. You can also rotate (and revoke) DSNs by navigating to [Project] > Settings > Client Keys (DSN).

I don't see this as something that should be configurable for any other customers. Forking and customizing is the preferred option in that case.

packages/map-template/src/index.jsx
markhougaard: @ammapspeople I think so. It may be worth seeing if we can make this more dynamic at some point, if the need arises, but I don't think that traffic is going to skew any data points in a significant way.

packages/map-template/src/index.jsx
markhougaard: @ammapspeople This stumped me as well, but I think I found out why we need it:

 Yesterday, when testing, I removed this part completely, given we don't use React Router. At that point, no data was sent to Sentry at all for Performance, only Error Logging. I think that there's a built-in mechanism in React that creates a single route for `/` that the Sentry SDK reads from, so without this part, nothing can be initialized and sent off.

Generally, with us having no routes defined, we need to be quite creative with our setup to be able to track properly at the Component level instead.

markhougaard:
> > > Does this work for the built Web component and React component? I see that a lot of configuration is done on the app level, but we only build and export the MapsIndoorsMap component. Is this configuration passed down?
> > 
> > 
> > Great question @ammapspeople. What would be the easiest way for me to test the React/Web Component locally to see if the data is actually sent off?
> 
> You can try to follow this guide: https://github.com/MapsPeople/web-ui/tree/main/packages/tests

Following the guide worked!

Moreover, I moved the Sentry initialization to `MapsIndoorsMap.jsx` based on your recommendation, @ammapspeople. Worked fine in my test 😀

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -111,6 +113,36 @@ function MapsIndoorsMap(props) {
     )
 }
 
+Sentry.init({
+    dsn: "https://0ee7fa162023d958c96db25e99c8ff6c@o351128.ingest.sentry.io/4506851619831808",
+    // Set environment to localhost if the url includes it. Otherwise, set to production.
+    environment: window.location.hostname === 'localhost' ? 'localhost' : 'production',
+    integrations: [
+        // See docs for support of different versions of variation of react router
```
ammapspeople, line 121: I still think a comment should be made about why we use React Router stuff here even though we don't use the React Router in the project otherwise.

packages/map-template/.gitignore
```diff
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+# Sentry Config File
+.env.sentry-build-plugin
```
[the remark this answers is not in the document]
markhougaard: @ammapspeople When running locally, yes, you will need to have a file like that with the relevant Sentry information in it. However, I've set them as secrets on this repo, so they should be read from there as part of the build and shipping of this app, and thus the file is not required for anything else than local testing.

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
markhougaard: @ammapspeople I added a comment in the code now 👍 (I thought you just wanted one in GitHub 🙃)

ammapspeople: Perfect 👍

ammapspeople (approved):
🆗 👌 🙆‍♂️

andreeaceachir142 (approved):

PR #344, opened 2024-03-11
Add map-template-release-cdn build step

markhougaard:
This adds a new job to the workflow file to release the built React and Web Components to a Google Cloud Storage Bucket.

It uploads the Components to two folders inside the bucket every time we create an npm release:

1. `latest`
2. `vX.Y.Z`, corresponding to the new release's version number.

Moreover, the cache is set to `0` for `latest`, and 5 minutes (`300`) for the version folders, so latest always serves the newest file, no matter what, while the version folders cache the files a bit longer.

Whether we just use the bucket name, or set up another url is out of scope for this PR.


ammapspeople:
By setting cache to `0` for the latest version, do we then risk any performance issues?

markhougaard:
I'm closing this off, as using jsDelivr for CDN, we seem to be able to do what we want. The branch should be kept around if we want to pick this up again, but for now, we'll try a different approach.

PR #345, opened 2024-03-13
Use jsDelivr as CDN for Map Template Web Component

markhougaard:
We could just go ahead and use jsDelivr as the CDN, but it caches even the `@latest` version very hard, and that would be a problem for us (it can take up 7 days before it's updated).

Including a workflow step to purge the CDN, we should see better results than using unpkg as the CDN, and hopefully don't run into trouble with files that are cached for too long.

### DIFF
--- a/packages/midt/package.json
+++ b/packages/midt/package.json
@@ -18,6 +18,7 @@
   "sass": "mixins/_index.scss",
   "style": "midt.css",
   "unpkg": "build/css/variables.css",
+  "jsdelivr": "build/css/variables.css",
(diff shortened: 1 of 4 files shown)


packages/midt/package.json
```diff
@@ -18,6 +18,7 @@
   "sass": "mixins/_index.scss",
   "style": "midt.css",
   "unpkg": "build/css/variables.css",
```
andreeaceachir142, line 20: Is this line still necessary with the `unpkg`?

PR #346, opened 2024-03-13
Set purge job as dependent on Map Template release

markhougaard:


markhougaard:
@andreeaceachir142 This PR fixes two things:

1. The "purge" job should only run after the Map Template Release was done (either skipped or completed)
2. The "MIDT" package was actually missing a `release.sh` file to be able to release. Goes to show it's been quite a while since we made an update to MIDT... The action that failed is here: https://github.com/MapsPeople/web-ui/actions/runs/8264797314/job/22609144690#step:11:10

andreeaceachir142 (approved):

PR #347, opened 2024-03-13
Add CDN-files

markhougaard:


ammapspeople (approved):
Very clever! 👏

markhougaard:
Tested this with @ammapspeople by my side, and it still works for older implementations that also define the Custom Element in inline-scripts.

It's marked as a minor update because you now don't have to define the Custom Element yourself.

PR #348, opened 2024-03-15
Add support for searching in the same venue by default

andreeaceachir142:
# What 

- Add support for searching in the same venue by default 
- Add new parameter called `searchAllVenues` which is a boolean that can be set in order to perform search across all venues in a solution 
- Fix issue with categories not updating when changing the venue

# How 

- Add new boolean parameter `searchAllVenues` and create an atom for storing the value
- Rename the `currentVenueName` for consistency reasons 
- In the `Search.jsx` file in the wrapper component, add a check if the `searchAllVenues` prop is true in the `mi-venue`
- In the `Search.jsx` file, add the venue attribute when searching for selecting a category in order to get the filtered locations 
- Add a new `useEffect()` on the `MapTemplate.jsx` file which handles the changes in the `currentVenueName` atom, and gets the current categories based on that

### DIFF
--- a/packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
+++ b/packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
@@ -4,8 +4,6 @@ import { useRecoilValue, useRecoilState } from 'recoil';
 import userPositionState from '../../../atoms/userPositionState';
 import languageState from '../../../atoms/languageState';
 import searchInputState from '../../../atoms/searchInputState';
-import kioskLocationState from '../../../atoms/kioskLocationState';
-import { useIsKioskContext } from '../../../hooks/useIsKioskContext';
 import currentVenueNameState from '../../../atoms/currentVenueNameState';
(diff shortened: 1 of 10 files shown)


ammapspeople:
Should this have a major label instead of minor?

andreeaceachir142:
That's up for us to decide 🤷🏻‍♀️ @ammapspeople 
I just added a minor for now, but we can change it if we decide that it's a major.

packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
```diff
@@ -44,6 +44,10 @@ const SearchField = forwardRef(({ placeholder, mapsindoors, results, clicked, cl
     /** Instruct the search field to search for Locations near the map center. */
     const searchNear = useNear();
 
+    const currentVenue = useRecoilValue(currentVenueNameState);
```
ammapspeople, line 47:
```suggestion
    const currentVenueName = useRecoilValue(currentVenueNameState);
```

ammapspeople (approved):
👏

PR #349, opened 2024-03-19
Indicate non-selectable locations in search results list

andreeaceachir142:
# What 

- Indicate non-selectable locations in search results list

# How

- As agreed with UX and Product team, the non-selectable locations in the search results list should not allow any interaction and they should be grayed out. 
- Add a `disabled` attribute to the non-selectable locations which is used to set the CSS styles

### DIFF
--- a/packages/map-template/src/components/WebComponentWrappers/ListItemLocation/ListItemLocation.jsx
+++ b/packages/map-template/src/components/WebComponentWrappers/ListItemLocation/ListItemLocation.jsx
@@ -2,6 +2,8 @@ import { useEffect, useRef } from 'react';
 import { useRecoilValue } from 'recoil';
 import mapsIndoorsInstanceState from '../../../atoms/mapsIndoorsInstanceState';
 import { useTranslation } from 'react-i18next';
+import './ListItemLocation.scss';
+
 
 /**
  * React wrapper around the custom element <mi-list-item-location>.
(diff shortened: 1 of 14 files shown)


packages/map-template/src/components/WebComponentWrappers/ListItemLocation/ListItemLocation.jsx
```diff
@@ -26,12 +27,17 @@ function ListItemLocation({ location, locationClicked, icon, isHovered }) {
...
-             // Check if the location is non-selectable before unhovering it
+            // Check if the location is non-selectable before unhovering it
             if (!location.properties.locationSettings?.selectable !== false) {
                 mapsIndoorsInstance.unhoverLocation(location);
             }
         }
 
+        // Add a disabled attribute to the non-selectable locations.
+        if (location.properties.locationSettings?.selectable === false) {
+            elementRef.current.setAttribute("disabled", "");
```
ammapspeople, line 38: Can we set a class name instead? The `disabled` attribute is usually used on form elements (inputs etc.). The `<mi-list-item-location>` element is not a form element, which may cause confusion.

andreeaceachir142: Good idea, replaced it with a class `non-selectable`.

ammapspeople (approved):
🎯

PR #350, opened 2024-03-21
Fix the bottom sheet jumping outside the viewport when selecting a category 

andreeaceachir142:
# What 

- Fix the bottom sheet jumping outside the viewport when selecting a category 

# How 

- Add a new atom indicating when the bottom sheet has been loaded, which is set when the transition has ended on the `Sheet.jsx` component 
- Add a new atom indicating if there is a category property/URL parameter, and based on that, call the `scrollIntoView` on the `Categories.jsx` file. The scroll into view should only happen when the category has been set as a prop or URL parameter, and **not** when the user selects a category.


packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -136,14 +141,14 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef }) {
...
 
     /*
-     * Use the active category state to scroll into view
-     * and center the category.
+     * If the active category is a prop/query parameter and the bottom sheet is loaded, 
+     * then scroll into view and center the active category. 
      */
     useEffect(() => {
-        if (activeCategory) {
+        if (activeCategory && hasCategoryProp && isBottomSheetLoaded) {
```
ammapspeople, line 148: Will this work on larger screens where the bottom sheet is not used?

andreeaceachir142: Good catch. I adjusted the check to also consider the desktop screens where there's no bottom sheet.

packages/map-template/src/atoms/hasCategoryPropState.js
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'recoil';
+
+const hasCategoryPropState = atom({
```
ammapspeople, line 3: Would it be more clean to have the value from the prop as the value in the atom? We already have atoms for properties (locationId, startZoomLevel etc.), and this boolean state is kinda off pattern-wise.

andreeaceachir142: Good idea. Refactored the category atom. ✅

ammapspeople (approved):
🆗

andreeaceachir142:
A fix for the issue reported in Jira has been merged in the latest commit 🙏🏻 @ammapspeople

PR #352, opened 2024-03-26
Add 2D/3D switch button

andreeaceachir142:
# What 

- Implement 2D/3D switch button 

# How 

- Add a new atom `is3DToggledState` which indicates if the 3D features are toggled 
- If the map is ready, check if the 3D features are toggled and then show the 3D/2D features according to the switch. Tilt the map to 45 degrees when showing the 3D features and tilt the map back to 0 degrees when showing 2D features.
- Create a `useState` which indicates if the switch component should be shown or not
- Add logic for showing the switch component if the `mapbox`, `3dwalls` and `floorplans` are enabled


packages/map-template/src/components/Switch/Switch.jsx
```diff
@@ -0,0 +1,37 @@
...
+import { ReactComponent as Light2D } from '../../assets/2d-light.svg';
+import { ReactComponent as Dark2D } from '../../assets/2d-dark.svg';
+import { ReactComponent as Light3D } from '../../assets/3d-light.svg';
+import { ReactComponent as Dark3D } from '../../assets/3d-dark.svg';
+import is3DToggledState from "../../atoms/is3DToggledState";
+
+/**
+ * Component responsible for switching between the 2D and 3D features.
+ */
+function Switch() {
```
ammapspeople, line 14: "Switch" is a very generic component name. Can we be more specific?

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
```diff
@@ -64,6 +68,24 @@ function MapboxMap({ onMapView, onPositionControl }) {
     }, []); // eslint-disable-line react-hooks/exhaustive-deps
     // We ignore eslint warnings about missing dependencies because onMapView should never change runtime and changing Mapbox Access Token runtime will give other problems.
 
+
+    // If the map is ready, check if the 3D features are toggled 
+    useEffect(() => {
+        if (mapView && mapView.isReady && isMapReady) {
+            // Check if the 3D features are switched on
+            // Conditionally render either 2D or 3D features depending on the Recoil Value of is3DToggled
```
ammapspeople, line 76:
```suggestion
            // Conditionally render either 2D or 3D features depending on the value of is3DToggled
```

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -481,6 +485,23 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         }
     }, [useKeyboard, kioskOriginLocationId, mapsindoorsSDKAvailable]);
 
+    /*
+     * React on changes to the solution.
+     * Decide whether to show the Switch component depending on the modules enabled on the solution.
+     */
+    useEffect(() => {
```
ammapspeople, line 492: Assuming this is related to Mapbox only, and not Google Maps, would it make sense to move all this into the MapboxMap component? Also the use of the `<Switch />` component?

packages/map-template/src/components/Switch/Switch.jsx
andreeaceachir142: I have brainstormed this quite a lot, and have some options: SwitchViewModes, SwitchMapFeatures, Switch2D3DFeatures. Since the file can't start with a number, 2D3DSwitch is not an option. 
Do you have any preference or any other suggestions? @ammapspeople

ammapspeople: ViewModeSwitch? VisibilitySwitch?

andreeaceachir142: Renamed component to VisibilitySwitch ✅

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Good idea, moved it to MapboxMap component ✅

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
```diff
@@ -64,6 +73,40 @@ function MapboxMap({ onMapView, onPositionControl }) {
...
 
+    // If the map is ready, check if the 3D features are toggled 
+    useEffect(() => {
+        if (mapView && mapView.isReady && isMapReady) {
+            // Check if the 3D features are switched on
+            // Conditionally render either 2D or 3D features depending on the value of is3DToggled
+            // Tilt the map to 45 degrees when the 3D features are switched on, and to 0 degrees when the 2D features are switched on.
+            if (is3DToggled) {
+                mapView.hideFeatures([mapView.MapboxFeatures.MODEL2D, mapView.MapboxFeatures.WALLS2D])
+                mapView.tilt(45, 2000);
```
ammapspeople, line 84: Does this collide with the `pitch` property? Or the `timeout` property?

packages/map-template/src/components/MapTemplate/MapTemplate.scss
```diff
@@ -20,6 +20,10 @@
         .venue-selector__button {
             display: block;
         }
+
+        .switch {
```
ammapspeople, line 24: Should this be in the MapboxMap.scss file instead?

ammapspeople: Ah, true. This is the "global show/hide" stuff 😄

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
andreeaceachir142: Good find @ammapspeople 
Pushed the new changes with handling the initial value of the pitch and the timeout property.

packages/map-template/src/components/VisibilitySwitch/VisibilitySwitch.jsx
```diff
@@ -0,0 +1,37 @@
...
+ * Component responsible for switching between the 2D and 3D features.
+ */
+function VisibilitySwitch() {
+    const primaryColor = useRecoilValue(primaryColorState);
+
+    const [is3DToggled, setIs3DToggled] = useRecoilState(is3DToggledState);
+
+    return (
+        <div className="switch">
+            <div className="switch__button"
```
ammapspeople, line 21: I think we should use `<button>` elements for buttons 😄

andreeaceachir142: Hmmmm, it's not really a button in its own definition. More like a switch element. 
What do you think? Should the class be changed or the tag?

ammapspeople: A switch contains clickable elements. Those can be anchor or button elements.
The floor selector can also be thought of as a switch conceptually, and those are implemented as `<button>` elements 😛 

Having them as actual buttons also improves accessibility, since you will be able to trigger them with the keyboard.

andreeaceachir142: Sounds fair, I replaced the `div` tags with `button` tags.

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
```diff
@@ -64,6 +72,50 @@ function MapboxMap({ onMapView, onPositionControl }) {
...
+    useEffect(() => {
+        if (solution) {
+            const isMapboxModuleEnabled = solution.modules.map(module => module.toLowerCase()).includes('mapbox');
+            const is3DWallsModuleEnabled = solution.modules.map(module => module.toLowerCase()).includes('3dwalls');
+            // The module floorplan refers to 2D Walls 
+            const is2DWallsModuleEnabled = solution.modules.map(module => module.toLowerCase()).includes('floorplan');
+
+            if (isMapboxModuleEnabled && is3DWallsModuleEnabled && is2DWallsModuleEnabled) {
+                setShowVisibilitySwitch(true);
+            }
```
ammapspeople, line 114: Can be shortened to 

```js
            // If all of the modules "mapbox", "3dwalls" and "floorplan" (2D walls) are enabled on the solution, show the VisibilitySwitch component
            if (['mapbox', '3dwalls', 'floorplan'].every(requiredMoule => solution.modules.map(module => module.toLowerCase()).includes(requiredMoule))) {
                setShowVisibilitySwitch(true);
            }
```
but I'm not sure if that is more readable.

andreeaceachir142: Yes, it would be! I made the change ✅

packages/map-template/src/components/Map/MapboxMap/MapboxMap.jsx
```diff
@@ -64,6 +72,46 @@ function MapboxMap({ onMapView, onPositionControl }) {
...
+    }, [mapView, is3DToggled, isMapReady, mapView?.isReady, pitch]);
+
+    /*
+     * React on changes to the solution.
+     * Decide whether to show the VisibilitySwitch component depending on the modules enabled on the solution.
+     */
+    useEffect(() => {
+        if (solution) {
+            // If all of the modules "mapbox", "3dwalls" and "floorplan" (2D walls) are enabled on the solution, show the VisibilitySwitch component
+            if (['mapbox', '3dwalls', 'floorplan'].every(requiredMoule => solution.modules.map(module => module.toLowerCase()).includes(requiredMoule))) {
```
ammapspeople, line 108:
```suggestion
            if (['mapbox', '3dwalls', 'floorplan'].every(requiredModule => solution.modules.map(module => module.toLowerCase()).includes(requiredModule))) {
```
🙃

andreeaceachir142: Oops 😂

PR #353, opened 2024-03-27
Add current venue name when venue has changed on map

andreeaceachir142:
# What 

- Fix the issue of current venue name not changing from the default venue 

# How 

- Changed the `AUSTIN` to the correct default venue `AUSTINOFFICE`
- Added the `setCurrentVenueName` when the venue is changed on the map


ammapspeople (approved):
👌

ammapspeople (approved):
👏

PR #354, opened 2024-04-08
Add workflow to name Sentry Releases with versions

markhougaard:
Sentry docs: https://docs.sentry.io/product/cli/releases/

Uses this GitHub Action: https://github.com/marketplace/actions/sentry-release


andreeaceachir142:
@markhougaard from what I know, when creating a react app, the source maps are generated under the folder `map-template/build/assets`. They are the `.map` files under that folder. 
I'm not 100& sure I'm correct here, so maybe @ammapspeople has some input too.

ammapspeople:
@markhougaard @andreeaceachir142  

- When building the React app for production, source map files are located alongside the distribution files in the `build/assets` folder.
- When building for npm, source map files are also located alongside the distribution files in the `dist` folder.

So if we can set a relative folder for the source map, it should work.

andreeaceachir142:
I think @markhougaard is planning to release this

PR #356, opened 2024-04-15
Fix QR code URL not working on the Google Cloud Storage buckets

andreeaceachir142:
# What 

- Remove extra "/" character when building the QR code URL

# How 

- When building the `targetUrl`, remove the extra "/" as it is unnecessary 
- Tested with on 3 different environments: localhost, map.mapsindoors.com and also storage.googleapis.com/mi-map-latest/index.html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23752,7 +23752,7 @@
     },
     "packages/map-template": {
       "name": "@mapsindoors/map-template",
-      "version": "1.43.0",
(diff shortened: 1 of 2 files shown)


ammapspeople (approved):
Do you remember if we wrote some test cases when we struggled with the URL building last time?

andreeaceachir142:
@ammapspeople Yess, I'll send that in a DM to you.

PR #359, opened 2024-05-15
Hide external ids by default

andreeaceachir142:
# What 

- Hide external IDs by default when performing a search in the Map Template

# How 

- Set the `show-external-ids` property to be false by default

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23769,7 +23769,7 @@
     },
     "packages/map-template": {
       "name": "@mapsindoors/map-template",
-      "version": "1.44.2",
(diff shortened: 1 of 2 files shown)


ammapspeople (approved):
I know it is agreed upon, but I'm afraid of some backlash, since some customers may rely on the external ID to be shown.

Ideally this should be configurable.

andreeaceachir142:
@ammapspeople We have added on the roadmap the implementation of a toggle/query parameter which in the future will allow the customers to configure whether they would like to show or hide the external ids.

PR #365, opened 2024-06-04
Make the exported Map Template React component testable before publishing

ammapspeople:
Expose a new React app that serves as a test app for the exported Map Template React Component.


andreeaceachir142:
Approved ✅ @ammapspeople I was able to run it on my machine.

andreeaceachir142 (approved):

PR #368, opened 2024-06-05
Fix Category scroll bar button

matbmapspeople:
What:
- fixing scss for scroll buttons
- fixing disabled state

How:
- for the style I added space-between
- for disabled state I found out that we are not reaching second `else if `. I studied a little those math operations and created one for `else if `. I just simply console logged all three: scrollWidth, scrollLeft and clientWidth. 
- I also used useEffect() for 'scroll' event.

Working solution:

https://github.com/MapsPeople/web-ui/assets/56337591/b7428422-4c50-4b90-9247-ed7c3241d1e4


andreeaceachir142:
@matbmapspeople Not sure what the changes in the Categories.jsx are for, the functionality works well when scrolling to the right, that the button becomes disabled. Test here: https://map.mapsindoors.com/

If we have less than 3 categories, the right button still appears to be clickable. 
image

matbmapspeople:
I was sure I fixed it.. I will look into it more @andreeaceachir142

andreeaceachir142 (approved):
👏

PR #371, opened 2024-06-12
Fix console error for clearing search input field

matbmapspeople:
I copied isNullOrUndefined function from the previous merge request and used it here as well.
We explicitly check is `keyboardRef.current` is not null but also if it is undefined it will crash anyway. 

I improved a check and everything seems to work.


andreeaceachir142 (approved):
Remember to merge in main and update the changelog with a patch version. ✅

PR #372, opened 2024-06-12
Fix kiosk modal wrapper

matbmapspeople:
word-break - prevents normal size words to be broken in half when there is a next line
overflow-wrap - ensures that if there is a very, very, very long word, it is going to be broken to the next line

Working example:
 
Screenshot 2024-06-12 at 10 03 27


andreeaceachir142 (approved):
Remember to merge in main and update the changelog with a patch. ✅

packages/map-template/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.47.1] - 2024-06-13
+
+### Fixed
+
+- Fixed an issue where the text in the Information Modal did not follow the expected formatting.
```
andreeaceachir142, line 12:
```suggestion
- Fixed an issue where the text in the Legend Information Modal in the Kiosk mode did not follow the expected formatting.
```

PR #375, opened 2024-06-13
Add support for Spanish language

andreeaceachir142:
# What 

- Add support for Spanish language in the Map Template

# How 

- Add new es.js file with all the Spanish translations
- Add Spanish layout for the keyboard


andreeaceachir142:
@matbmapspeople When reviewing, see example here: https://github.com/MapsPeople/web-ui/pull/373

PR #379, opened 2024-06-18
Don't use the whole IETF language string when setting the MapsIndoors SDK language

ammapspeople:
**What**

Don't use the whole IETF language string when setting the MapsIndoors SDK language.

**Why**

The SDK/backend does not work properly when using the whole IETF language string. Eg. using `es-co` will cause the fetched data to fallback to the default language, since it is only possible to define translations for `es` in the MapsIndoors platform.

**How**

Only use the first part of the language string, eg. `es` or `en` when calling `setLanguage()`. Now when eg. using a browser with the language set to `es-co`, it will instruct the SDK to set the language as `es`, which corresponds better with the data.


andreeaceachir142:
Test OK ✅ @ammapspeople
Ready for release

PR #381, opened 2024-06-20
Create params for skipGo and accessibility

matbmapspeople:
What: 
- Introducing new skipGo URL param - when present in the URL, the route should appear straightway without a need to click 'Go' button
- Introducing new accessibility URL param - when present in the URL, accessibility should be taken into consideration when getting a route

How:
- Introducing webcomponent.js, queryParameter and setters for both: skipGo and accessibility.
- If condition for accessibility is `true`, then we assign **checked** to `true`.
- If condition for skipGo is true, then we call `onStartDirections` function which normally is called when we click 'Go' button.
- I created `isFinishRouteState` atom, so when we click 'Finish Route', we are not ending up in infinite loop of rendering the route.

Working solution:

Uploading Screen Recording 2024-06-20 at 11.22.55.mov…

### DIFF
--- a/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
+++ b/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
@@ -497,12 +497,12 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
                         {travelMode === travelModes.DRIVING && <DriveIcon />}
                         {travelMode === travelModes.BICYCLING && <BikeIcon />}
                         <div>{t('Distance')}:</div>
-                        <div className="wayfinding__meters">{totalDistance && <mi-distance unit={distanceUnitSystem} meters={totalDistance} />}</div>
+                        <div className="wayfinding__meters">{totalDistance && isDirectionReady && <mi-distance unit={distanceUnitSystem} meters={totalDistance} />}</div>
                     </div>
                     <div className="wayfinding__time">
                         <ClockIcon />
                         <div>{t('Estimated time')}:</div>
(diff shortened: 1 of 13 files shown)


matbmapspeople:
I will take care of read me and changelog later. I want to make sure that logic is correct @andreeaceachir142

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -387,6 +396,15 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
 
+    useEffect(() => {
+        if (skipGo && isSet && !isFinishRoute) {
+            if (directionsFromLocation && directionsToLocation && isDirectionReady && !isSkipGoUsed) {
+                onStartDirections();
+                setIsSkipGoUsed(true)
+            }
+        }
+    }, [skipGo, isSet, directionsFromLocation, directionsToLocation, isDirectionReady])
+
```
andreeaceachir142, line 407: Add some docs on what is happening here

packages/map-template/src/hooks/useDirectionsInfo.js
```diff
@@ -3,24 +3,28 @@ import getLocationPoint from '../helpers/GetLocationPoint';
...
 
     useEffect(() => {
+        setIsDirectionReady(false);
         if (originLocation?.geometry && destinationLocation?.geometry) {
             directionsService.getRoute({
                 origin: getLocationPoint(originLocation),
                 destination: getLocationPoint(destinationLocation),
                 travelMode: travelMode,
-                avoidStairs: accessibilityOn
+                avoidStairs: accessibilityOn || accessibility[0]
```
andreeaceachir142, line 27: Use recoil value instead

ammapspeople:
@matbmapspeople Did I understand it correctly, that we've decided no to do this for the main branch? If so, this pull request could be closed.

PR #386, opened 2024-08-06
Fix scroll buttons disabled/enabled states in the Kiosk mode

matbmapspeople:
- I updated `updateScrollButtonsState()` function so it gets a proper disabled state when locations' length is lower than 8. Number 8 indicates maximum number of locations that are visible in the element without scroll.

- Inside `Search.jsx`, I called newly created function `newCategorySelected()`. When new category is selected, and all locations to that category are rendered, we need to update disabled/enabled state of scroll buttons.

- I performed a lot of scouting.


packages/map-template/src/components/Search/Search.jsx
```diff
@@ -132,6 +132,11 @@ function Search({ onSetSize, isOpen }) {
         setSearchResults(locations);
         setFilteredLocations(locations);
         setShowNotFoundMessage(locations.length === 0);
+
+        // When category changes, we need to update scroll buttons enabled/disabled states.
```
ammapspeople, line 136: I think readers deserve to know why we grab hold of the `transitionend` event instead of doing it right away.

matbmapspeople: ✅

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -140,6 +140,15 @@ function Search({ onSetSize, isOpen }) {
         setSearchResults(locations);
         setFilteredLocations(locations);
         setShowNotFoundMessage(locations.length === 0);
+
+        // Handles updates to scroll buttons when the category changes.
+        // When a category changes, the scroll buttons need to have their enabled/disabled states updated.
+        // Since some categories might load before the SCSS element is fully rendered, we listen for the 'transitionend' event.
```
ammapspeople, line 146: "SCSS element"? 🤔 Did you mean DOM element?

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -5,12 +5,32 @@ import { Component, h, JSX, Method, Prop, Watch } from '@stencil/core';
...
+    @Prop() locations;
+
+    /**
+     * Method.
+     */
+    @Method()
+    public async newCategorySelected(locations): Promise<any> {
+        this.locations = locations;
+        this.updateScrollButtonsState();
+    }
```
ammapspeople, line 29: The scroll buttons element is not tightly coupled with Locations and Categories, but can control any type of content. Please make the naming more generic.

packages/map-template/src/components/Search/Search.jsx
matbmapspeople: ✅

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
matbmapspeople: ✅

ammapspeople:
There's still some naming referring to locations and categories in the scroll buttons web component.

matbmapspeople:
> There's still some naming referring to locations and categories in the scroll buttons web component.

How about now @ammapspeople

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -5,12 +5,32 @@ import { Component, h, JSX, Method, Prop, Watch } from '@stencil/core';
...
      */
     @Prop() scrollContainerElementRef: HTMLDivElement;
+
+    /**
+     * Items that are rendered inside a specific element.
+     */
+    @Prop() elementItems;
+
+    /**
+     * Method.
```
ammapspeople, line 23: 😂

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -5,12 +5,32 @@ import { Component, h, JSX, Method, Prop, Watch } from '@stencil/core';
...
+     */
+    @Method()
+    public async updateKioskScrollButtons(elementItems): Promise<any> {
```
ammapspeople, line 26: The phrase "kiosk" should not be here.

packages/components/src/components.d.ts
```diff
@@ -853,6 +853,14 @@ export namespace Components {
...
     }
     interface MiScrollButtons {
+        /**
+          * Locations visible when specific category is selected.
+         */
+        "locations": any;
+        /**
+          * Method.
+         */
+        "newCategorySelected": (locations: any) => Promise<any>;
```
ammapspeople, line 863: Remember to run the build command to update the declaration file.

matbmapspeople:
New changes pushed @ammapspeople

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -140,6 +140,15 @@ function Search({ onSetSize, isOpen }) {
...
         setShowNotFoundMessage(locations.length === 0);
+
+        // Handles updates to scroll buttons when the category changes.
+        // When a category changes, the scroll buttons need to have their enabled/disabled states updated.
+        // Since some categories might load before the DOM element is fully rendered, we listen for the 'transitionend' event.
+        // The 'transitionend' event is triggered when the DOM element changes its size, which can occur as a result of new categories being fetched.
+        // Upon completion of the size transition, the 'updateKioskScrollButtons' function is triggered to handle the updated state.
+        searchRef.current?.addEventListener('transitionend', () => {
+            scrollButtonsRef?.current?.updateKioskScrollButtons(locations);
+        });
```
ammapspeople, line 151: Still references kiosk phrasing on the scroll buttons component.

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -5,12 +5,32 @@ import { Component, h, JSX, Method, Prop, Watch } from '@stencil/core';
...
      * Reference to the element with scroll on parent element.
+     *
      * @type {HTMLDivElement}
      */
     @Prop() scrollContainerElementRef: HTMLDivElement;
+
+    /**
+     * Items that are rendered inside a specific element.
+     */
+    @Prop() elementItems;
```
ammapspeople, line 20: I don't see elementItems being set from outside of the component. If this is internal only, it should use a `@state` decorator.

ammapspeople: Also, is it even needed? Can't we calculate the number of child elements when we need it?

matbmapspeople: I used `@State`. I was thinking to keep it as separate variable, so it is easier to understand what it is. @ammapspeople

ammapspeople: But is it needed? We are forcing the consumer of the component to add additional info to the method call updateScrollButtons. It seems as the scroll buttons component should already have access to it (using some variant of querySelector, or by other means).

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -63,17 +88,26 @@ export class ScrollButtons {
         } else if (this.upButtonElement.disabled) {
             this.upButtonElement.disabled = false;
         }
-
         // Disable or enable the scroll down button
-        if (this.scrollContainerElementRef.scrollHeight - this.scrollContainerElementRef.scrollTop === this.scrollContainerElementRef.clientHeight) {
-            this.downButtonElement.disabled = true;
-        } else if (this.downButtonElement.disabled) {
+        // length 8 is just that maxiumum amount of element items visible without a scroll possibility
```
ammapspeople, line 92: But what if the element container is of a different size? This number 8 seems tied to the usage in the Map Template. This component should be able to work in other places too.

matbmapspeople:
I cleaned up a little bit now and we do not use additional property. We also do not check for length. @ammapspeople

matbmapspeople:
What do you think about it now @ammapspeople ☝️

packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -5,12 +5,26 @@ import { Component, h, JSX, Method, Prop, Watch } from '@stencil/core';
...
 export class ScrollButtons {
     /**
      * Reference to the element with scroll on parent element.
+     *
      * @type {HTMLDivElement}
      */
     @Prop() scrollContainerElementRef: HTMLDivElement;
+
+    /**
+     * Updates scroll buttons enabled/disabled states when elementItems amount changes.
```
ammapspeople, line 18: This sounds like it happens automatically ("when elementItems amount changes"), but it is a handle to manually trigger recalculation of buttons state if needed. Please rephrase so it doesn't sound like stuff happens automatically.

matbmapspeople: ✅

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -140,6 +140,15 @@ function Search({ onSetSize, isOpen }) {
...
         setShowNotFoundMessage(locations.length === 0);
+
+        // Handles updates to scroll buttons when the category changes.
+        // When a category changes, the scroll buttons need to have their enabled/disabled states updated.
+        // Since some categories might load before the DOM element is fully rendered, we listen for the 'transitionend' event.
+        // The 'transitionend' event is triggered when the DOM element changes its size, which can occur as a result of new categories being fetched.
+        // Upon completion of the size transition, the 'updateScrollButtons' function is triggered to handle the updated state.
+        searchRef.current?.addEventListener('transitionend', () => {
+            scrollButtonsRef?.current?.updateScrollButtons();
+        });
```
ammapspeople, line 151: Some cleanup and performance fixes are needed here.

- Don't apply the event listener if it is not needed (wrap in a check for `isKioskContext`).
- The `updateScrollMethods` method is triggered many, many times, and the event listener is not cleared up, causing memory leaks to happen. Consider applying a `{ once: true }` (but do check if that actually works) to the event listener or any other means to reduce memory leaks.

matbmapspeople: Wrapped in `isKioskContext`
And added `once: true` - tested and seems to work. @ammapspeople

ammapspeople (approved):

PR #387, opened 2024-08-14
Fix bug where pitch property didnt work when 2d/3d switch was not shown

ammapspeople:
# What

- Fix bug where pitch property was not respected in case 2D/3D switch was not activated.

# How

- Make sure always to default to the set pitch state, instead of undefined in case there is no Mapbox view mode.


andreeaceachir142 (approved):
Approved ✅

PR #389, opened 2024-09-12
Bugfix for wayfinding skipping first step

ammapspeople:
When setting a route on the DirectionsRenderer, wait for setRoute to resolve before setting the step index. The setRoute method recently became asynchronous, meaning that the step index would be reset to -1 just before resolving, causing a mismatch in which step is active.


andreeaceachir142 (approved):
✅

PR #390, opened 2024-09-16
Add placeholder property to mi-dropdown

matbmapspeople:
For PI improvement, I need a different text inside placeholder. To not break current interface, I made this implementation.


matbmapspeople:
Do I need to change anything else to make it available to CMS components after releasing it to PROD? @ammapspeople

packages/components/src/components/dropdown/dropdown.tsx
```diff
@@ -99,6 +99,14 @@ export class Dropdown {
      */
     @Prop() filterable = false;
 
+
+    /**
+     * If present, it dictates placeholder for an filterable input field in the dropdown. Defaults to 'Type to filter...'.
+     *
+     * @type {string}
+     */
+    @Prop() placeholder: string;
```
ammapspeople, line 108: https://stenciljs.com/docs/properties#default-values

Looks like we can do 

```suggestion
    @Prop() placeholder: string = 'Type to filter...';
```

(not tested)

ammapspeople:
> Do I need to change anything else to make it available to CMS components after releasing it to PROD? @ammapspeople

You'll need to update the components package in the CMS to the version in which this will be released.

packages/components/src/components/dropdown/dropdown.tsx
matbmapspeople: Tested, works. Suggestion applied.

ammapspeople (approved):
LGTM 😁

PR #395, opened 2024-10-02
Map as standalone react component  basic map

ammapspeople:
# What

- Create a new repo package, "react-components", containing a Vite+React setup made the same way as the Map Template.
- Implement a `MIMap` component with two sub components `MapboxMap` and `GoogleMapsMap`.
  - The map will take MapsIndoors API key, map provider token and a center+zoom level and show a MapsIndoors map based on that.


packages/react-components/src/App.jsx
```diff
@@ -0,0 +1,22 @@
...
+        <>
+            <h1>MapsIndoors React Components</h1>
+            <h2><code>&lt;MIMap /&gt;</code></h2>
+            <div style={{ marginLeft: '5vw', width: "90vw", height: "70vh", boxShadow: '0px 0px 20px #000' }}>
+                <MIMap
+                    apiKey="mapspeople3d"
+                    gmApiKey={import.meta.env.VITE_GOOGLE_MAPS_API_KEY}
+                    mapboxAccessToken={import.meta.env.VITE_MAPBOX_ACCESS_TOKEN}
+                    center={{ lat: 57.05825378109134, lng: 9.951020621279326 }}
+                    zoom={19.2}
```
andreeaceachir142, line 15: Is the zoom 19,2 just a hard coded value?

packages/react-components/src/components/MIMap/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -0,0 +1,59 @@
...
+        if (!mapViewInstance) return;
+        if (zoom) {
+            mapViewInstance.getMap().setZoom(zoom);
+        }
+    }, [zoom]);
+
+    useEffect(() => {
+        const loader = new GoogleMapsApiLoader({
+            apiKey: apiKey,
+            version: 'quarterly',
```
andreeaceachir142, line 35: Just for my own sake, what does "quarterly" mean in this context?

packages/react-components/src/components/MIMap/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -0,0 +1,59 @@
...
+            libraries: ['geometry', 'places']
+        });
+
+        loader.load().then(() => {
+            // Initialize Google Maps MapView
+            const mapViewOptions = {
+                element: document.getElementById('map'),
+                disableDefaultUI: true, // Disable Map Type control, Street view control and Zoom controls.
+                center: center ?? { lat: 0, lng: 0 }, // The MapsIndoors SDK needs a starting point and a zoom level to avoid timing issues when setting a venue.
+                zoom: zoom ?? 15
```
andreeaceachir142, line 45: Is the 15 the default zoom level we want Google Maps to show?

packages/react-components/src/components/MIMap/GoogleMapsMap/GoogleMapsMap.jsx
```diff
@@ -0,0 +1,59 @@
...
+            };
+
+            const mv = new window.mapsindoors.mapView.GoogleMapsView(mapViewOptions);
```
andreeaceachir142, line 48: Could we rename this to mapView instead, or keep mv but write a comment about it?

packages/react-components/src/components/MIMap/MIMap.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+MIMap.propTypes = {
+    apiKey: PropTypes.string.isRequired,
+    gmApiKey: PropTypes.string,
+    mapboxAccessToken: PropTypes.string,
+    center: PropTypes.object,
+    zoom: PropTypes.number
+}
+
+function MIMap({ apiKey, gmApiKey, mapboxAccessToken, center, zoom }) {
```
andreeaceachir142, line 19: Should we add some definitions to the function and all the parameters here?
Same for the Google and Mapbox maps...

packages/react-components/src/components/MIMap/MIMap.jsx
```diff
@@ -0,0 +1,66 @@
+import { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import MapboxMap from './MapboxMap/MapboxMap';
+import GoogleMapsMap from './GoogleMapsMap/GoogleMapsMap';
+
+const mapTypes = {
+    Google: 'google',
+    Mapbox: 'mapbox'
+};
```
andreeaceachir142, line 9: Would it make sense to move this outside as a constant? We might need to use it along the way, at least as far as I remember, we did that in the previous implementation.

packages/react-components/src/App.jsx
ammapspeople: As we discussed in the office, `App.jsx` serves as a test app for the `MIMap` component. The prop values here are for testing purposes.

packages/react-components/src/components/MIMap/GoogleMapsMap/GoogleMapsMap.jsx
ammapspeople: Read more about Google Maps JS API versioning here: https://developers.google.com/maps/documentation/javascript/versions#release-channels-and-version-numbers

packages/react-components/src/components/MIMap/GoogleMapsMap/GoogleMapsMap.jsx
ammapspeople: I can't remember where I have 15 from, but it cannot be 0 due to that causing glitches. I see that the Map Template map uses 21, so I'll use that here as well.

packages/react-components/src/components/MIMap/GoogleMapsMap/GoogleMapsMap.jsx
ammapspeople: There will be a `mapView` as part of a useState 🙃 I'll think of a better name.

packages/react-components/src/components/MIMap/MIMap.jsx
ammapspeople: I'll do it when we need it in more places 😄

andreeaceachir142 (approved):
Thanks for all the clarifications. Approved ✅

PR #396, opened 2024-10-17
Mitigate memory leaks

ammapspeople:
Mitigate memory leaks, primarily by cleaning up component state holding objects when unmounting.

Also updated Sentry to the latest version since that have had some past memory leaks as well.


andreeaceachir142 (approved):
👏

PR #398, opened 2024-10-17
Implement new prop for searching for external results

andreeaceachir142:
- Implement new prop `searchExternalResults` which determines if the search results in the Wayfinding view should include Mapbox/Google places or if they should only show MapsIndoors results. 
- By default it is set to false
- It is only the Wayfinding view that takes the new prop into consideration, and not the main Search view
- The prop is also taken into account when generating the QR code URL


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -544,6 +547,15 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setShowRoadNames(showRoadNames);
     }, [showRoadNames])
 
+    /*
+     * React on changes to the searchExternalResults prop.
+     */
+    useEffect(() => {
+        if (mapsindoorsSDKAvailable && searchExternalResults) {
```
ammapspeople: Is the check for SDK availability needed?

andreeaceachir142: Actually not, removed it ✅

ammapspeople (approved):
⌨️🔥!!!

PR #400, opened 2024-10-18
Map as standalone component: Expose MapsIndoors instance

ammapspeople:
Expose the MapsIndoors instance as a function prop and showcase that it works in the test app by setting up a MI click event listener.

### DIFF
--- a/packages/react-components/src/App.jsx
+++ b/packages/react-components/src/App.jsx
@@ -1,7 +1,22 @@
+import { useEffect, useState } from 'react';
 import './App.css'
 import MIMap from './components/MIMap/MIMap'
 
 function App() {
+
+    const [mapsIndoorsInstance, setMapsIndoorsInstance] = useState();
+
+    useEffect(() => {
+        if (mapsIndoorsInstance) {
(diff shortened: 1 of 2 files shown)


packages/react-components/src/App.jsx
```diff
@@ -1,7 +1,22 @@
...
+    const [mapsIndoorsInstance, setMapsIndoorsInstance] = useState();
+
+    useEffect(() => {
+        if (mapsIndoorsInstance) {
+            // For demo purpose: setup MapsIndoors click event listener that sets clicked location to be invisible.
+            mapsIndoorsInstance.on('click', (location) => {
+                const displayRule = mapsIndoorsInstance.getDisplayRule(location);
+                displayRule.visible = false;
+                mapsIndoorsInstance.setDisplayRule(location.id, displayRule);
+            });
```
andreeaceachir142, line 16: What was the use case for setting the clicked location to be invisible?

ammapspeople: Showcase. Demo. Nothing but that. It's not part of the Map component, but the test/demo app.

andreeaceachir142 (approved):

PR #401, opened 2024-10-18
Simplify README files

andreeaceachir142:
Cleaned up the README file which is displayed on the GitHub page and removed all the duplicate code for props and query parameters. With this change, we avoid maintaining two separate README files. 
Instead, a link to NPM has been shared which can be accessed to see all the props and query parameters.


ammapspeople (approved):
🥳 🎉 👍 🥇 🏆 🍹

PR #409, opened 2024-11-19
Demo for offices release 1: setup

ammapspeople:
- Create a new package, `demo-office`, with a React app created with Vite
- Use the MIMap component in the app. Very unfinished state 😄 

FYI: I have created a separate ticket to fix an unwanted dependency of `vite-plugin-svgr` for the `MIMap` component. We cannot expect everyone to add that in order to use the component.


andreeaceachir142 (approved):
🆗 🆗 🆗

PR #410, opened 2024-11-19
Limit the number of search results displayed in the Search box

Lucaci-Andrei:
## What
- Limit the amount of results shown in the Search bar
- Limit the amount of results shown in the Search bar when searching by category

## How
- Create a `MAX_RESULTS` constant that holds the maximum amount of results to show

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24020,7 +24020,7 @@
     },
     "packages/map-template": {
       "name": "@mapsindoors/map-template",
-      "version": "1.56.0",
+      "version": "1.58.0",
       "devDependencies": {
         "@googlemaps/js-api-loader": "^1.15.1",
         "@mapsindoors/components": "*",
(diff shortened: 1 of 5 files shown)


andreeaceachir142:
@ammapspeople Looking through the documentation of the `mi-search` component, we have an attribute called `miTake`: https://components.mapsindoors.com/search/
Do we have a preference over which approach to take here for limiting the search results?

andreeaceachir142 (approved):

packages/map-template/src/components/WebComponentWrappers/Search/Search.jsx
```diff
@@ -21,8 +21,9 @@ import searchAllVenuesState from '../../../atoms/searchAllVenues';
  * @param {boolean} props.preventFocus - If set to true, the search field will be disabled.
  * @param {boolean} props.google - Set to true to include results from Google Places autocomplete service.
  * @param {boolean} props.mapbox - Set to true to include results from Mapbox Places autocomplete service.
+ * @param {number} amountOfSearchResults - The amount of search results to return.
```
ammapspeople, line 24:
```suggestion
 * @param {number} props.amountOfSearchResults - The amount of search results to return.
```

The props object is destructured.

ammapspeople (approved):
👏

PR #411, opened 2024-11-20
Demo for offices release 1: app shell

ammapspeople:
- A "fake" app shell to mimic that the map is part of a larger web app


packages/demo-office/src/App.css
```diff
@@ -1,12 +1,23 @@
...
-  height: 100vh;
-  width: 100vw;
-  display: grid;
-  grid-template-rows: 100px 1fr;
+    width: 100%;
+    height: 100dvh;
+    display: grid;
+    grid-template-rows: 56px 1fr;
+    background-color: #f1f5f9;
+    color: black;
```
andreeaceachir142, line 7: Can we use midt for the color here and for the px as well?

packages/demo-office/src/App.css
```diff
@@ -1,12 +1,23 @@
...
+        padding: 0 16px 32px 16px;
+    }
 }
 
 .mapsindoors-map {
-  height: 100%;
-  width: 100%;
+    height: 100%;
+    width: 100%;
+    border-radius: 8px;
```
andreeaceachir142, line 22: See if we can use midt for the px values here 🙌

andreeaceachir142 (approved):

PR #412, opened 2024-11-21
Demo for offices release 1: fake device position

ammapspeople:
# What

- Implement a "fake" blue dot on the Office demo

# How

- Adds a fake data set that contains a faked device position
- Adds an image created with the Canvas API (copied more or less from the old Vercel demo)
- Adds the faked device position as a data source on the Mapbox map
- Adds a new layer on the Mapbox map using the added data source and the image
- Remove the built-in Position Control from the MIMap since we don't want multiple blue dots


andreeaceachir142 (approved):
🆗 🆗 🆗

PR #414, opened 2024-11-26
fix: Correct venue sorting to use display name instead of administrat…

Lucaci-Andrei:
## What
- Sort venues by display name
## How
- Use venue display name (`venue.venueInfo.name`) to sort venues alphabetically


ammapspeople (approved):
🎉

PR #415, opened 2024-11-27
Map as standalone component: Fix broken tilt when resetting

ammapspeople:
Make pitch/tilt work properly when resetting (ie. when using the timeout prop). This is reverting an earlier (wrong) decision to include the current pitch when fitting bounds.

However, it causes the same logic (which modules need to be enabled in order to show the viewmode switch) to appear in both the react-components package and now also the map-template package. Any ideas to mitigate this @andreeaceachir142 ?

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -139,7 +139,7 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
     const [initialFilteredLocations, setInitialFilteredLocations] = useState();
 
     const [appConfig, setAppConfig] = useRecoilState(appConfigState);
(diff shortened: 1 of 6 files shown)


andreeaceachir142:
Are you referring to the two files `MapTemplate` and `ViewModeSwitch`?  @ammapspeople

ammapspeople:
> Are you referring to the two files `MapTemplate` and `ViewModeSwitch`? @ammapspeople

Yes. Both need to know if the view mode switch is "on", and as early as possible.

ammapspeople:
@andreeaceachir142 As discussed IRL, the MIMap now communicates to the parent component if the view mode switch is shown or not, and the Map Template can now use that information to do what it needs to do instead of running the same check. Ready for re-review 😄

ammapspeople:
I also refactored how the View mode switch is reset to use the same reset count pattern that we use in the Map Template to reset the UI.

PR #417, opened 2024-11-28
Map as standalone component: Fix timing issue in ViewModeSwitch

ammapspeople:
fix: wrap the view mode handling in a promise to make sure the MapView is ready. Because if it is not, the calls to hideFeatures will not work (current SDK bug).


andreeaceachir142 (approved):
Ship ship ship ✅🎉

PR #419, opened 2024-11-29
Map as a standalone component: make sure a venue is set

ammapspeople:
Fix: make sure a venue will always be selected in order to ensure that 2D geometry is always loaded.

This offloads the burden of having the consumer of the component needing to do this (setting a venue).


andreeaceachir142 (approved):
🆗

PR #420, opened 2024-12-02
Update UI on the wayfinding page

andreeaceachir142:
# What 

- Update UI on the wayfinding page

# How 

- Update the text on the wayfinding page and the placeholder text for the input fields. 
- Update the length of the input fields to only cover 90% 
- Update all translations with the new strings

### DIFF
--- a/packages/map-template/src/components/Search/Search.scss
+++ b/packages/map-template/src/components/Search/Search.scss
@@ -43,12 +43,13 @@
         background-color: white;
     }
 
+    &__label {
+        span {
+            display: none;
+        }
+    }
+
     @media (max-width: $desktop-breakpoint) {
         padding: var(--spacing-medium) var(--spacing-x-small) var(--spacing-medium) var(--spacing-x-small);
     }
-
-    span {
-        display: none;
-    }
-
 }
--- a/packages/map-template/src/components/Search/Search.scss
+++ b/packages/map-template/src/components/Search/Search.scss
@@ -47,4 +47,8 @@
         padding: var(--spacing-medium) var(--spacing-x-small) var(--spacing-medium) var(--spacing-x-small);
     }
 
+    span {
+        display: none;
+    }
+
 }
--- a/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
+++ b/packages/map-template/src/components/Wayfinding/Wayfinding.jsx
@@ -409,7 +409,7 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
                     <CloseIcon />
                 </button>
                 <div className="wayfinding__locations">
-                    <label>
+                    <label className="wayfinding__label">
                         <span>{t('Choose departure')}</span>
                         <SearchField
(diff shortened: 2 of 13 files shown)


andreeaceachir142:
@Lucaci-Andrei just for reference, this is an example of ticket where the strings needed to be translated in all the supported languages.

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -402,47 +402,40 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
-                            mapbox={selectedMapType === mapTypes.Mapbox && searchExternalLocations}
-                            placeholder={t('Search by name, category, building...')}
-                            results={locations => searchResultsReceived(locations, searchFieldIdentifiers.FROM)}
-                            clicked={() => onSearchClicked(searchFieldIdentifiers.FROM)}
-                            cleared={() => onSearchCleared(searchFieldIdentifiers.FROM)}
-                            changed={() => onInputChanged(searchFieldIdentifiers.FROM)}
-                        />
-                    </label>
+                    <SearchField
+                        ref={fromFieldRef}
```
ammapspeople, line 413: Now that the explicit label was removed, consider adding an `aria-label` attribute on the inputs.

packages/map-template/src/components/Wayfinding/Wayfinding.scss
```diff
@@ -207,4 +199,8 @@
         background: var(--color-gray-10);
         border: 1px solid var(--color-gray-30);
     }
+
+    mi-search{
```
ammapspeople, line 203:
```suggestion
    mi-search {
```

packages/map-template/src/components/Wayfinding/Wayfinding.scss
```diff
@@ -207,4 +199,8 @@
...
+        width: 90%;
```
ammapspeople, line 204: Why exactly 90%? Would it be more exact to use a `calc(100% - "whatever it needs to not cover ...px");` ? 😄

andreeaceachir142: Good idea, updated it ✅

andreeaceachir142:
Ready for re review ✅ @ammapspeople

packages/map-template/src/components/Search/Search.scss
```diff
@@ -47,4 +47,8 @@
         padding: var(--spacing-medium) var(--spacing-x-small) var(--spacing-medium) var(--spacing-x-small);
     }
 
+    span {
```
ammapspeople, line 50: Consider targeting using a BEM element selector (`&__label`? 🤔 ), just to avoid problems down the line if we add a `<span>` in another place in the component.

andreeaceachir142: Good point ✅

ammapspeople (approved):
🆗 x 3

PR #421, opened 2024-12-02
Map as standalone react component

ammapspeople:
# What

- Create a new package, `react-components`, which contains a `MIMap` component which wraps the basic map functionality.
- Use the `MIMap` component in the Map Template package.


ammapspeople:
Closing due to hard-to-solve dependency conflicts. Will be re-done in another branch.

PR #423, opened 2024-12-10
Release 1.70.0

Lucaci-Andrei:
- Added `MAX_RESULTS` constant to limit the maximum number of search results displayed at once 
- Sort venues by display name


ammapspeople:
@Lucaci-Andrei Should we make it 1.61.0 instead of skipping a bunch of minor versions? 😁

PR #428, opened 2024-12-10
Make google maps work again

ammapspeople:
Avoid early exit in case the solution is not set. The solution is only set when the map is a Mapbox map, and this caused the viewModeSwitchVisible not to be set for Google maps, causing the map never to declare itself initialized.


andreeaceachir142 (approved):
🆗 🆗 🆗

PR #431, opened 2024-12-13
Map template: use the new goto method

ammapspeople:
# Why

We don't want to zoom in unreasonably far on the map when choosing Locations (clicking on search results, setting as prop) with small geographical size.

# How

The MapsIndoors JS SDK was recently released with a new `goTo` method. We use that now to pan the map initially and when clicking on a Location in a search result list.

We thus restrict the map from zooming in further than level 22 (MapsIndoors zoom level).

This made the existing `goto*` methods in the `useMapBoundsDeterminer` hook obsolete, so they were removed.


packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -92,7 +87,7 @@ const useMapBoundsDeterminer = () => {
 
                         getDesktopPaddingBottom().then(desktopPaddingBottom => {
                             setMapPositionKnown(kioskLocation.geometry);
-                            goToGeometry(mapType, kioskLocation.geometry, mapsIndoorsInstance, desktopPaddingBottom, 0, startZoomLevel, currentPitch, bearing);
+                            goto(kioskLocation.geometry, mapsIndoorsInstance, desktopPaddingBottom, 0, startZoomLevel, currentPitch, bearing);
```
andreeaceachir142, line 90: Should probably follow the casing to be `goTo` instead. 😄

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -152,11 +147,9 @@ const useMapBoundsDeterminer = () => {
 
 export default useMapBoundsDeterminer;
 
-
 /**
  * Make the map go to specified geometry with optional padding, zoom level, pitch and bearing.
  *
- * @param {string} mapType
  * @param {object} geometry
```
andreeaceachir142, line 153: Missing zoom level from the docs here.
Also, should we write some docs for each of there params?

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -165,82 +158,15 @@ export default useMapBoundsDeterminer;
  * @param {number} pitch
  * @param {number} bearing
  */
-function goToGeometry(mapType, geometry, mapsIndoorsInstance, paddingBottom, paddingLeft, zoomLevel, pitch, bearing) {
```
andreeaceachir142, line 168: Why has the mapType been removed here?

ammapspeople: The information is no longer needed, since it's the SDK that takes care of knowing which map provider to send instructions to.

packages/map-template/src/hooks/useMapBoundsDeterminer.js
ammapspeople: 🫡

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -182,8 +182,8 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
 
             const miSdkApiTag = document.createElement('script');
             miSdkApiTag.setAttribute('type', 'text/javascript');
-            miSdkApiTag.setAttribute('src', 'https://app.mapsindoors.com/mapsindoors/js/sdk/4.38.0/mapsindoors-4.38.0.js.gz');
-            miSdkApiTag.setAttribute('integrity', 'sha384-gbCO1T6ALl7CofEGTlZ1Sk0UfyNoEem6oADmOdFVW4Fe0sS1N63u/Yqht83LEb20');
+            miSdkApiTag.setAttribute('src', 'http://app.mapsindoors.com/mapsindoors/js/sdk/4.38.1/mapsindoors-4.38.1.js.gz');
```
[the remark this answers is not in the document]
ammapspeople: ✅

packages/map-template/src/hooks/useMapBoundsDeterminer.js
ammapspeople: ZoomLevel is on line 157.
I (or someone not human) will add some descriptions.

andreeaceachir142 (approved):
🆗 🆗 🆗

PR #432, opened 2024-12-13
Dropdown enhancements

eNepper:
I made a few minor changes to the dropdown component.
- Added an alignment attribute to set the dropdown on which side of the button will align. Possible values are left and right, and the default is left. 
- If set the label will be shown as the value in the dropdown button. (Not only for multi-select)

### DIFF
--- a/packages/components/src/components/dropdown/dropdown.tsx
+++ b/packages/components/src/components/dropdown/dropdown.tsx
@@ -575,17 +575,21 @@ export class Dropdown {
         return noResultsTemplate;
     }
 
-
     /**
      * Helper function to render the button label.
      *
      * @returns {JSX.Element}
      */
     renderButtonLabel(): JSX.Element {
-        const label = this.label ?? this.selected?.[0]?.text ?? this.selected?.[0]?.innerHTML ?? '';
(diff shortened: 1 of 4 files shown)


packages/components/src/components/dropdown/dropdown.tsx
```diff
@@ -575,17 +582,10 @@ export class Dropdown {
...
     renderButtonLabel(): JSX.Element {
+        const label = this.label ?? this.selected?.[0]?.text ?? this.selected?.[0]?.innerHTML ?? '';
         this.selected?.[0]?.text ?? this.selected?.[0]?.innerHTML ?? '';
 
-        if (this.multiple) {
-            return (<div part="button-label" class="button__label">{this.label}</div>);
-        } else if (this.selected?.[0]?.text) {
-            return (<div part="button-label" class="button__label">{this.selected[0].text}</div>);
-        } else if (this.selected?.[0]?.innerHTML) {
-            return (<div part="button-label" class="button__label button__label--from-inner-html" innerHTML={this.selected[0].innerHTML}></div>);
```
ammapspeople, line 585: Should we somehow preserve the styling given from the `--from-inner-html` modifier?

ammapspeople (approved):
🆗

PR #435, opened 2025-01-09
Click on non MapsIndoors data to exit categories if active.

Lucaci-Andrei:
### What
---
- Fix issue of clicking outside of MapsIndoors data to close the Categories tab

### How
---
 - Implement the usage of the custom hook useOutsideMapsIndoorsDataClick in Categories
 - Add a prop to Categories that determines if it is open or not

### DIFF
--- a/packages/map-template/src/components/Search/Categories/Categories.jsx
+++ b/packages/map-template/src/components/Search/Categories/Categories.jsx
@@ -24,7 +24,7 @@ import mapsIndoorsInstanceState from "../../../atoms/mapsIndoorsInstanceState";
  * @param {function} props.onSetSize - Callback that is fired when the categories are clicked.
  * @param {function} props.getFilteredLocations - Function that gets the filtered locations based on the category selected.
  * @param {object} props.searchFieldRef - The reference to the search input field.
- * @param {boolean} props.isOpen - Whether the Categories window is open or not
+ * @param {boolean} props.isOpen - Determines wheteher the Categories window is open or not.
  */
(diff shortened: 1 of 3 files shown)


packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -124,6 +130,20 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef }) {
...
+            setFilteredLocations([]);
+            setSize(snapPoints.FIT);
+
+            // If search field has value, trigger search without category filter
+            if (searchFieldRef.current?.getValue()) {
+                searchFieldRef.current.triggerSearch();
+            }
+        }
+    }, [clickedOutsideMapsIndoorsData]);
+
```
andreeaceachir142, line 146: Could we write up some comments on this useEffect on what it does?

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -22,8 +24,9 @@ import categoryState from "../../../atoms/categoryState";
  * @param {function} props.onSetSize - Callback that is fired when the categories are clicked.
  * @param {function} props.getFilteredLocations - Function that gets the filtered locations based on the category selected.
  * @param {object} props.searchFieldRef - The reference to the search input field.
+ * @param {boolean} props.isOpen - Whether the Categories window is open or not
```
andreeaceachir142, line 27:
```suggestion
 * @param {boolean} props.isOpen - Determines wheteher the Categories window is open or not.
```

andreeaceachir142 (approved):

PR #436, opened 2025-01-09
Add center prop and logic around it

matbmapspeople:
I added possibility to read _center_'s value from the URL.

Based on this value and other URL prop values I created a logic where the map should be centred.


andreeaceachir142:
@ammapspeople I've added a few comments here which stood out at a first glance, to save you some time 😄

ammapspeople (commented):
I'll pause reviewing this until the build error mentioned in the Jira ticket is resolved.

matbmapspeople:
I now addressed all your comments @andreeaceachir142 and @ammapspeople

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -72,51 +80,132 @@ const useMapBoundsDeterminer = () => {
...
+            // Returns Venue that intersects with center prop.
+            const intersectingVenueWithCenterPoint = venuesInSolution.find(venue => {
+                return turf.booleanIntersects(venue.geometry, centerPoint.geometry);
+            })
+
+            // If startZoomLevel is not defined, fallback to 16 default value.
+            const zoomLevel = startZoomLevel ?? 16;
+
+
+            Promise.all([getDesktopPaddingBottom(), getDesktopPaddingLeft(), getMobilePaddingBottom()]).then(([desktopPaddingBottom, desktopPaddingLeft, mobilePaddingBottom]) => {
```
ammapspeople, line 107: Won't we risk that the "pad" variables are null since this is asynchronous and the code following it is immediately executed?

ammapspeople:
The bounds are not fit to venue when no center prop provided:

image

ammapspeople:
The center prop is not reacting to changes runtime. It should.

https://github.com/user-attachments/assets/a2e00a87-dac8-4e41-8eed-b60cbd307e45

matbmapspeople:
I believe I addressed all your comments @ammapspeople
I would appreciate @andreeaceachir142 if you can resolve yours if I managed to fix your issues :)

matbmapspeople:
I applied your suggestions @ammapspeople

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -62,61 +70,157 @@ const useMapBoundsDeterminer = () => {
...
 
+            // Parse center prop into coordinates. If it is not included in the URL, latLng are undefined.
+            const [latitude, longitude] = center
+                ? center.split(",").map(Number)
+                : [undefined, undefined];
+
+            // Create centerPoint object.
+            const centerPoint = { geometry: { type: 'Point', coordinates: [latitude, longitude] } };
+
+            // If startZoomLevel is not defined, fallback to 16 default value.
```
ammapspeople, line 95: This is written like it's a universal thing, but it's only related to when the center prop is set, right?

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -62,61 +70,157 @@ const useMapBoundsDeterminer = () => {
...
         if (mapsIndoorsInstance && currentVenue) {
             setMapPositionInvestigating(true);
 
+            // Parse center prop into coordinates. If it is not included in the URL, latLng are undefined.
+            const [latitude, longitude] = center
+                ? center.split(",").map(Number)
+                : [undefined, undefined];
+
+            // Create centerPoint object.
+            const centerPoint = { geometry: { type: 'Point', coordinates: [latitude, longitude] } };
```
ammapspeople, line 93: Find a way to only make these operations when we know they are needed.

ammapspeople: Pull it out into a separate function maybe.

matbmapspeople: I created a helper function that seems to make code cleaner.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -92,8 +93,9 @@ defineCustomElements();
  * @param {boolean} [props.showRoadNames] - A boolean parameter that dictates whether Mapbox road names should be shown. By default, Mapbox road names are hidden when MapsIndoors data is shown. It is dictated by `mi-transition-level` which default value is 17.
  * @param {boolean} [props.showExternalIDs] - Determine whether the location details on the map should have an external ID visible. The default value is set to false.
  * @param {boolean} [props.searchExternalLocations] - If you want to perform search for external locations in the Wayfinding mode. If set to true, Mapbox/Google places will be displayed depending on the Map Provider you are using. If set to false, the results returned will only be MapsIndoors results. The default is true.
+ * @param {string} [props.center] - If you want the map to show a specific coordinates, provide the center value, f.ex.: center=9.9,57.05 (latitude, longitude). If the specified coordinates intersect with a Venue, that Venue will be selected in the Venue Selector.
```
ammapspeople, line 96: It's not just about the Venue selector, but also eg. what is searchable 😁

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -62,61 +70,149 @@ const useMapBoundsDeterminer = () => {
...
      * determine where to make the map go to.
      */
     function determineMapBounds() {
         const currentVenue = venuesInSolution.find(venue => venue.name.toLowerCase() === currentVenueName.toLowerCase());
+
         if (mapsIndoorsInstance && currentVenue) {
             setMapPositionInvestigating(true);
 
+            // If center prop is defined and startZoomLevel is not defined, fallback to 18 default value.
+            const zoomLevel = startZoomLevel ?? 18;
```
ammapspeople, line 88: The variable name sounds more global than it is.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
matbmapspeople: ✅

packages/map-template/src/hooks/useMapBoundsDeterminer.js
matbmapspeople: ✅

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
ammapspeople: https://github.com/MapsPeople/web-ui/pull/436/commits/ef705ffd60e0429071f0f663d4f0270267a8eff6 - why not use what was already committed?

matbmapspeople: I updated `MapTemplate` and `MapsIndoorsMap` files.

packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -62,61 +70,149 @@ const useMapBoundsDeterminer = () => {
...
      * determine where to make the map go to.
      */
     function determineMapBounds() {
         const currentVenue = venuesInSolution.find(venue => venue.name.toLowerCase() === currentVenueName.toLowerCase());
+
         if (mapsIndoorsInstance && currentVenue) {
             setMapPositionInvestigating(true);
 
+            // If center prop is defined and startZoomLevel is not defined, fallback to 18 default value.
+            const initialZoomLevel = startZoomLevel ?? 18;
```
ammapspeople, line 88: I maintain that this is too "globally" named and probably also not in the right place.
To avoid confusion when people read this, consider the following:

- Just use `startZoomLevel ?? 18` in the few places it's needed
- Maybe even add a variable for storing the 18 value so we only need to change it in one place if needed
- Create a function for returning the `startZoomLevel ?? 18` value like the getCenterPoint

matbmapspeople: I added helper function.

I also added a check in `useCurrentVenue.js` file, because in some circumstances, `...mainmenu.find(...` was undefined.

@ammapspeople ☝️

ammapspeople: @matbmapspeople Looks like you've then solved issue number 577 in our internal Jira tool. Remember to update it accordingly.

ammapspeople:
It now "lands" on unexpected map bounds when no props are set but the default ones.

I suspect the eager use of the `getZoomLevel` function instead of `startZoomLevel` value, but I may be wrong. Please asses every use of that function if it's meant to be used.

image

matbmapspeople:
Fixed @ammapspeople

PR #438, opened 2025-01-09
Map template: fit bounds when selecting category

ammapspeople:
# What

Make the map fit to matching Locations when the user clicks/taps on a Category.

# How

- A new function was created in the Search component, `fitMapBoundsToLocations`. When called, it will build an array of GeoJSON features to be passed to the MapsIndoors JS SDK method `goTo` which will ensure that the map is fitted to encompass those Locations.
- The `onResults` function in the the Search component was extended with an optional `fitMapBounds` parameter, which when set to true will call the new `fitMapBoundsToLocations` function.
- This `fitMapBounds` parameter is set to true only when calling from the `getFilteredLocations` function, which is only run when the user clicks on a Category.
- Also fixed a bug causing a media query match always to be false if called early. This caused the fitting to category locations when using the category prop to use the mobile padding on desktop sized screens since the value of `isDesktop` would be falsely false (😁) at that time.


packages/map-template/src/components/Search/Search.jsx
```diff
@@ -157,6 +162,46 @@ function Search({ onSetSize, isOpen }) {
...
+    function fitMapBoundsToLocations(locations) {
+        if (!mapsIndoorsInstance.goTo) return; // Early exit to prevent crashes if using an older version of the MapsIndoors JS SDK. The goTo method was introduced in version 4.38.0.
+
+        const currentFloorIndex = mapsIndoorsInstance.getFloor();
+
+        // Create a GeoJSON FeatureCollection from the locations that can be used as input to the goTo method.
+        const featureCollection = {
+            type: 'FeatureCollection',
+            features: locations
+                // Filter out locations that are not on the current floor. Including those could result in a wrong fit since they are not visible on the map anyawy.
```
matbmapspeople, line 181:
```suggestion
                // Filter out locations that are not on the current floor. Including those that could result in a wrong fit since they are not visible on the map anyway.
```

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -157,6 +162,46 @@ function Search({ onSetSize, isOpen }) {
...
+                .filter(location => parseInt(location.properties.floor, 10) === parseInt(currentFloorIndex, 10))
+
+                // Filter out locations that are not part of the current venue. Including those when fitting to bounds could cause the map to zoom out too much.
+                .filter(location => location.properties.venueId.toLowerCase() === currentVenueName.toLowerCase())
```
matbmapspeople, line 185: Is it correct that we compare venueId and venue name? Should this be .venueId === currentVenueId (or something like that)?

ammapspeople: currentVenueName is the administrative ID (we call it both "name" and "administrative ID"  for ... reasons, for example in the CMS), and is represented by `venueId` in the Location properties.

matbmapspeople: Got it ✅

matbmapspeople:
LGTM ✅

Lucaci-Andrei:
👍

andreeaceachir142 (approved):
🆗 🆗 🆗

PR #441, opened 2025-01-14
Fix miTransitionLevel using recoil

matbmapspeople:
I removed using atoms inside MapboxMap.jsx
I made it so miTransitionLevel is respected in MapboxMap.jsx, if defined in the URL.


matbmapspeople:
Suggestion applied @ammapspeople

packages/react-components/src/components/MIMap/MapboxMap/MapboxMap.jsx
```diff
@@ -21,7 +19,7 @@ MapboxMap.propTypes = {
     resetViewMode: PropTypes.number,
     mapsIndoorsInstance: PropTypes.object,
     viewModeSwitchVisible: PropTypes.bool,
-    mapOptions: PropTypes.object
+    mapOptions: PropTypes.object,
```
ammapspeople, line 22:
```suggestion
    mapOptions: PropTypes.object
```

packages/map-template/src/components/MapWrapper/MapWrapper.jsx
```diff
@@ -274,13 +277,21 @@ function MapWrapper({ onLocationClick, onMapPositionKnown, useMapProviderModule,
         onTileStyleChanged(mapsIndoorsInstance);
     }, [tileStyle]);
 
+    /**
+     * React on changes in the miTransitionLevel prop.
+     */
+    useEffect(() => {
+        setMapOptions({ miTransitionLevel: miTransitionLevel})
```
ammapspeople, line 284: To avoid the unexpected, maybe try to only set the property and value of the miTransitionLevel is a number (or can be parsed to a number).

matbmapspeople: Done ✅

ammapspeople (approved):
🆗 🆗 🆗

matbmapspeople:
@ammapspeople I pushed a small fix that @andreeaceachir142 found.

PR #452, opened 2025-01-28
Restructure location details to respect React best practices

Lucaci-Andrei:
# What 
- Restructure locationDetails component to follow React best practices
- Identify functions that could use React's useCallback to improve performance
# How
- Move useEffects() before return statement
- Convert the back function to a useCallback


ammapspeople (approved):
🥳

PR #455, opened 2025-01-30
Fix too early zooming when using startZoomLevel

matbmapspeople:
There was a conflict while merging previous bugfixes. 

`mapPositionKnown` is not passed to any goTo() function, which is wrong.

I added `mapPositionKnown` to all places needed.


ammapspeople (approved):
👍

PR #458, opened 2025-02-03
Fix: goTo() function refactoring

matbmapspeople:
goTo will not setZoom level, since it fits to bounds
When startZoomLevel is not 18, and is defined in the URL, setCenter to be center of a geometry with a specific zoom level

### DIFF
--- a/packages/map-template/src/hooks/useMapBoundsDeterminer.js
+++ b/packages/map-template/src/hooks/useMapBoundsDeterminer.js
@@ -282,10 +282,10 @@ function goTo(geometry, mapsIndoorsInstance, paddingBottom, paddingLeft, zoomLev
             padding: { top: 0, right: 0, bottom: paddingBottom, left: paddingLeft },
         })
     } else {
-        const centerOfBoundingBox = turf.center(geometry);
+        const centerOfGeometry = turf.center(geometry);
         const mapView = mapsIndoorsInstance.getMapView();
 
-        mapView.setCenter({ lat: centerOfBoundingBox.geometry.coordinates[1], lng: centerOfBoundingBox.geometry.coordinates[0]} )
+        mapView.setCenter({ lat: centerOfGeometry.geometry.coordinates[1], lng: centerOfGeometry.geometry.coordinates[0]} )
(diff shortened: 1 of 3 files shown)


packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -270,17 +268,24 @@ export default useMapBoundsDeterminer;
...
-    mapsIndoorsInstance.goTo({ type: 'Feature', geometry, properties: {} }, {
-        maxZoom: zoomLevel ?? 22,
-        padding: { top: 0, right: 0, bottom: paddingBottom, left: paddingLeft },
-    }).then(() => {
-        if (zoomLevel && mapPositionKnown !== false) {
-            mapsIndoorsInstance.setZoom(zoomLevel);
-        }
-    });
+function goTo(geometry, mapsIndoorsInstance, paddingBottom, paddingLeft, zoomLevel, pitch, bearing) {
+    const defaultZoomLevel = 18;
```
ammapspeople, line 273: What does default zoom level mean? Where does the value come from?

matbmapspeople: @ammapspeople  This value comes from getZoomLevel() function when default zoom level is defined as 18.

ammapspeople: Seems like knowledge in two places 😁 

Also the Map Template does not have a general/public concept of a default zoom level. Can we strengthen the naming here to narrow it down to when it is used?

matbmapspeople: How about `initialLoadZoomLevel` @ammapspeople? 
This constant has a value that indicates a zoom level that we would experience on initial load/default load.

matbmapspeople:
It has been tested okay by @Lucaci-Andrei. 
Can I get a final approval for it @Lucaci-Andrei?

Lucaci-Andrei:
Looks good to me 👍 . Tested and approved 🤝

PR #459, opened 2025-02-04
Feature: share button in location details

ammapspeople:
# What

Adds share functionality in the Location Details, making it possible for the user to copy a link to the clipboard or scan a QR code. The link will open the Map Template with current Location as a query parameter.

# How

- Extract logic for generating the base link for sharing into a Recoil Selector, `baseLink`.
- Refactor the `QRCodeDialog` component to be more general instead of coupled to directions link.
- Replaced the `showQRCodeDialogState` boolean to a `qrCodeLinkState` containing the link, which also acts as a trigger to show the QR code dialog.
- In the `Directions` component, set the share directions link using the `qrCodeLinkState`, thus triggering the QR Code Dialog. 
- Implement a new component, `ShareLocationLink`, with buttons for copying a link to a clipboard or activating the QR code dialog.
	- The component uses a new NPM package, `react-click-away-listener` to help detecting when the user clicks outside the ShareLocationLink UI - in that case, we want to close it.
	- The QR code button is newer shown on small (mobile) screens.
- Update QR code icon as per wishes from UX


andreeaceachir142:
Really nice and clean implementation @ammapspeople 🙌

andreeaceachir142 (approved):

Lucaci-Andrei (approved):

Lucaci-Andrei:
Looks good to me 👍

matbmapspeople:
All good from my side ✅ @ammapspeople

PR #466, opened 2025-02-12
Fix 2D/3D switch not respecting primaryColor prop

Lucaci-Andrei:
# What 
- Fix 2D/3D selector not respecting the primaryColor query parameter
# How
- Refactor mapOptions by moving the declaration up the dependency tree
- Add a handle function that can be passed down to child components that can add additional values to mapOptions
- Set up primary color prop in mapOptions


ammapspeople (approved):
👏

matbmapspeople (approved):

PR #467, opened 2025-02-14
In wayfinding: make it possible to click on map to set origin or destination

ammapspeople:
# What

When in wayfinding, make it possible to choose a from or to location by clicking on a MapsIndoors location on the map.

- When the "from" field is active, clicking on a MapsIndoors location on the map will use that Location as the origin no matter if another was always selected
- When the "to" field is active, clicking on a MapsIndoors location on the map will use that Location as the destination only if the "to" field is empty.
- 

# How

- Created a new atom, `wayfindingLocation`, which is used to instruct the Wayfinding component to use the stored location as either origin or destination.
- Defined three different actions of what can happen when clicking on a MapsIndoors location on the map.
- In the MapTemplate component: 
	- Refactored the `locationClicked` method and the useEffect for `currentMapView` to use the new atom and the action types.
	- Abolished the `locationsDisabledRef` along the way :)
- In the Wayfinding component:
	- React to changes in the `wayfindingLocation` atom in order to populate the "from" or the "to" field and subsequently trigger the route request.

### DIFF
--- a/packages/map-template/CHANGELOG.md
+++ b/packages/map-template/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.71.0] - 2025-02-26
+
+### Added
+
+- In Wayfinding view, is is now possible to click on a MapsIndoors Location on the map to set it as either origin or destination, depending on which field is active (note: to set destination, the destination field must be empty)
+
 ## [1.70.0] - 2025-02-20
 
 ### Changed


matbmapspeople:
When selecting 'from' location and navigating to another location in another Venue we do not have results (which is fine). After clearing the wayFinding box and going back to initial venue - wayFinding does not work for a route that normally works:
Let me know if you need more explanation @ammapspeople 😄 

https://github.com/user-attachments/assets/f93d02ee-4ede-4341-aef3-b2ac09f1c169

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -335,6 +337,33 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
...
+            return;
+        }
+
+        // Trigger route search by resetting all these:
+        setSearchResults([]);
+        setSearchTriggered(false);
+        setHasGooglePlaces(false);
+        setShowMyPositionOption(false);
+        setWayfindingLocation(null); // will re-trigger another run with early exit but necessary in order to be able to re-click already clicked locations on the map
+    }, [wayfindingLocation]);
```
Lucaci-Andrei, line 365: Could we add these into a function and call it here 😄 ?

ammapspeople:
> When selecting 'from' location and navigating to another location in another Venue we do not have results (which is fine). After clearing the wayFinding box and going back to initial venue - wayFinding does not work for a route that normally works: Let me know if you need more explanation @ammapspeople 😄

@matbmapspeople I have addressed the issue in the latest commit. Can you confirm that it works as expected for you now?

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
ammapspeople: Yes. Done 😄

matbmapspeople:
Nice @ammapspeople !!! I could not reproduce it anymore

matbmapspeople (approved):
LGTM ✅

andreeaceachir142 (approved):
Very nice implementation!  ✅🙌

PR #469, opened 2025-02-18
Fix wrong map bounds when using locationId or kioskOriginLocationId props

ammapspeople:
# What

Fixes bug where wrong map bounds are used when using the `locationId` or `kioskOriginLocationId` props. In those cases, the map should zoom to fit the location.

# How

Introduce the usage of an undefined zoom level when calling `goTo` for the mentioned cases. No zoom level should be used there.


matbmapspeople (approved):
LGTM ✅ 

It also has been tested and indeed I'm zoomed in when `locationId` or `kioskOriginLocationId` is defined in the URL.

PR #471, opened 2025-02-18
Feature: Listen for App Tile from App Config

matbmapspeople:
Listen for App Tile from App Config, otherwise fallback to the hardcoded, original one.

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -619,6 +619,13 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
     }, [center]);
 
     /**
+     * Sets app title based on app config.
+     */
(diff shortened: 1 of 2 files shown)


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -618,6 +618,13 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setCenter(center);
     }, [center]);
 
+    /**
+     * Sets app title based on app config.
+     */
+    useEffect(() => {
+        document.title = appConfig?.appSettings?.title ?? 'MapsIndoors Web';
```
ammapspeople, line 625: I disagree that this should be done at all.

What if a customer is using the Map Template web component as part of a larger web app app? Then we come along and change the title, and there's nothing they can do about it.

I believe the manipulation of the title tag belongs in the hosted web app and only there.

PR #472, opened 2025-02-18
Bugfix: Switch selection pin when route order is switching

matbmapspeople:
Working solution: 

https://github.com/user-attachments/assets/fd958a44-d1b5-4990-a039-18f2967ce7fc


packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -314,6 +317,13 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
             }
             setDestinationLocation(originLocation);
             setOriginLocation(destinationLocation);
+
+            // When switching route order from A->B to B->A, remove selection pin and select new originLocation.
+            mapsIndoorsInstance.deselectLocation();
+
+            if (mapsIndoorsInstance.deselectLocation) {
```
ammapspeople, line 324: Why check for the existence of a method that you just called? 😛

PR #473, opened 2025-02-19
Reposition zoom control and adjust margin in MapboxMap component

Lucaci-Andrei:


matbmapspeople (approved):
Code-wise it looks good. Have not tested it, but it can be pushed to release branch that later will be tested as one big chunk.

PR #475, opened 2025-02-20
Rearrange placement for venue selector and 2D/3D switch

Lucaci-Andrei:


matbmapspeople:
And here is the case when 2D/3D Switch is visible but Venue Selector not:

Screenshot 2025-02-21 at 12 49 09

ammapspeople:
The implementation breaks the layout in the standalone MIMap component and the Office Demo:

image

image

image

Lucaci-Andrei:
> The implementation breaks the layout in the standalone MIMap component and the Office Demo:
> 
> image
> 
> image
> 
> image

Fixed 👍

Lucaci-Andrei:
> When there is only one Venue - I assume we do not show Venue selector. In this case with have a funny looking UI:
> 
> Screenshot 2025-02-21 at 12 46 02
> 
> I guess we can check for `venues.length` for a specific solution and then apply a desired styling. Also, 2D/3D Switch is not visible here as well:

Fixed 👍

PR #476, opened 2025-02-21
Refactor/rearange UI elements  expandable categories

matbmapspeople:


packages/map-template/src/components/Search/Categories/Categories.scss
```diff
@@ -1,25 +1,55 @@
...
+                border: 1px solid var(--color-gray-30);
+                background: none;
+                cursor: pointer;
+                text-align: left;
+            }
+
+            img {
+                width: 24px;
+                height: 24px;
+                margin-right: 10px;
```
Lucaci-Andrei, line 32: We should use var(--spacing-large) for width/height instead of hard-coded values and maybe var(--spacing-small) for margin-right.

Same concept should be applied across they style document

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -109,33 +102,14 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
      * Update the state of the left and right scroll buttons
      */
     function updateScrollButtonsState() {
-        const { scrollLeft, scrollWidth, clientWidth } = categoriesListRef?.current || {};
+        const { scrollLeft } = categoriesListRef?.current || {};
```
Lucaci-Andrei, line 105: We may not need this at all , since we do not have any scrollLeft 🤔

PR #480, opened 2025-02-28
Don't snap to full height when entering wayfinding

ammapspeople:
# What

- Don't expand the bottom sheet to full height when entering Wayfinding.

# Why

- We want the mobile users to be able to click the map to set origin or destination. Having the bottom sheet in full height prevents that.

# How

- Set snap point to FIT instead of MAX.
- Created a new prop on the Sheet component instructing it to set up a Mutation Observer to listen for content changes when the Sheet is snapped to "FIT".
    - This required changing the contentHeight state to a ref to keep track inside the Mutation Observer.
- Fix wrong min height on Wayfinding.


packages/map-template/src/components/BottomSheet/BottomSheet.jsx
```diff
@@ -140,9 +140,10 @@ function BottomSheet({ directionsFromLocation, directionsToLocation, pushAppView
             />
         </Sheet>,
         <Sheet
-            minHeight="238"
+            minHeight="190"
```
andreeaceachir142, line 143: Is 190px the initial value of the bottom sheet in all the different views for example wayfinding, search etc?

packages/map-template/src/components/BottomSheet/Sheet/Sheet.jsx
```diff
@@ -51,25 +52,48 @@ function Sheet({ children, isOpen, minHeight, preferredSizeSnapPoint, onSwipedTo
...
+        fitMutationObserver = new MutationObserver(() => {
+            const height = contentRef.current.children.item(0).clientHeight;
+            contentHeightRef.current = height;
+            changeSheetHeight(snapPoints.FIT);
+        });
+
+        fitMutationObserver.observe(contentRef.current, {
+            childList: true,
+            subtree: true
+        });
```
andreeaceachir142, line 81: It might be beneficial to write what the childList and subtree are used for in this context here 🙏

packages/map-template/src/components/BottomSheet/BottomSheet.jsx
ammapspeople: No, just the initial value for this sheet that wraps the wayfinding component.

andreeaceachir142 (approved):
✅

matbmapspeople:
Is it expected to have so much white space at the bottom?

Screenshot 2025-03-04 at 08 55 57

matbmapspeople:
> > @ammapspeople I managed to reproduce it. First selection works as expected. As soon as I close it and try to select two locations again, I see a white box:
> 
> @matbmapspeople I think I fixed it in the latest change. Can I get you to confirm?

Yes, I can confirm it is fixed. Nice! 🎉 @ammapspeople

matbmapspeople (approved):
Code-wise and what I could test, everything looks good ✅

andreeaceachir142 (approved):

PR #481, opened 2025-03-03
Bugfix/make show road name work again

matbmapspeople:


packages/map-template/src/components/MapWrapper/MapWrapper.jsx
```diff
@@ -299,6 +301,13 @@ function MapWrapper({ onLocationClick, onMapPositionKnown, useMapProviderModule,
         }
     }, [miTransitionLevel])
 
+    /**
+     * React on changes in the showRoadNames prop.
+     */
+    useEffect(() => {
+        onMapOptionsChange({ showRoadNames: showRoadNames })
+    }, [showRoadNames])
+
```
andreeaceachir142, line 310: Is there no need to check if the showRoadNames prop actually exists?

matbmapspeople: No, because `showRoadNames` can be undefined. 
If we make a check here, if `showRoadNames` is undefined, we will not pass any value to the SDK.

andreeaceachir142 (approved):
✅

PR #482, opened 2025-03-04
Feature improve minZoom

matbmapspeople:
How: 
1. I set overall minZoom to be 10.
2. On Directions, when the polyline is visible, I set minZoom to null for both Mapbox and Google Maps. When route is finished, I set it back to 10.


matbmapspeople:
Closing this merge request. Got feedback from Anders how to improve it.

PR #485, opened 2025-03-06
Feature: Extend minZoom level and set minZoom level to null when wayFinding

matbmapspeople:
How?
- Inside `MIMap.jsx`, I removed code responsible for setting minZoom to 14.

- Inside `Directions.jsx`, I set minZoom to `null` for both Mapbox and Google Maps (depends which one is used by current Map Template). Then I added logic to `goTo()` current venue if we cancel wayfinding. 

- Inside `MapTemplate.jsx`, I added logic to set minZoom to 14.

### DIFF
--- a/packages/map-template/src/components/Directions/Directions.jsx
+++ b/packages/map-template/src/components/Directions/Directions.jsx
@@ -131,11 +131,7 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped, onRouteFinishe
                     setDestinationDisplayRule(mapsIndoorsInstance.getDisplayRule(directions.destinationLocation));
                 }
 
-                if (mapType === 'mapbox') {
-                    mapsIndoorsInstance.getMapView().getMap().setMinZoom(null);
-                } else if (mapType === 'google') {
-                    mapsIndoorsInstance.getMapView().getMap().setOptions({ minZoom: null })
-                }
+                setMinZoom(mapType, null);
             });
         }
     }, [isOpen, directions, mapsIndoorsInstance, travelMode]);
@@ -184,7 +180,7 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped, onRouteFinishe
     useEffect(() => {
         if (!isOpen && directionsRenderer) {
             stopRendering();
-            setMinZoom(mapType);
+            setMinZoom(mapType, 14);
         }
     }, [isOpen]);
 
@@ -246,12 +242,13 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped, onRouteFinishe
      * Sets minZoom for a specific map provider.
      * 
      * @param {string} mapType 
+     * @param {number} zoomLevel 
      */
-    function setMinZoom(mapType) {
+    function setMinZoom(mapType, zoomLevel) {
         if (mapType === 'mapbox') {
-            mapsIndoorsInstance.getMapView().getMap().setMinZoom(14);
+            mapsIndoorsInstance.getMapView().getMap().setMinZoom(zoomLevel);
         } else if (mapType === 'google') {
-            mapsIndoorsInstance.getMapView().getMap().setOptions({ minZoom: 14 })
+            mapsIndoorsInstance.getMapView().getMap().setOptions({ minZoom: zoomLevel })
         }
     }
 
--- a/packages/map-template/src/components/Directions/Directions.jsx
+++ b/packages/map-template/src/components/Directions/Directions.jsx
@@ -25,6 +25,7 @@ import showExternalIDsState from '../../atoms/showExternalIDsState';
 import PropTypes from "prop-types";
 import baseLinkSelector from '../../selectors/baseLink';
 import mapTypeState from "../../atoms/mapTypeState";
+import { ZoomLevel } from "../../constants/zoomLevelValues";
 
 let directionsRenderer;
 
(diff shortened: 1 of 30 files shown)


ammapspeople:
The changes to the `useReset` hook breaks the timeout functionality. The map must return to the initial venue when the timeout kicks in using the `useReset` hook.

Why are the changes to the `useReset` hook necessary in order to remove the minZoom when in Wayfinding mode? I seems like we are bundling an unrelated fix into this pull request.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -239,7 +246,8 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
     const handleMapOptionsChange = (newMapOptions) => {
         setMapOptions(previousMapOptions => ({
             ...previousMapOptions,
-            ...newMapOptions
+            ...newMapOptions,
+            minZoom: 14
```
andreeaceachir142, line 250: Do we still need the minZoom here?

matbmapspeople:
> The changes to the `useReset` hook breaks the timeout functionality. The map must return to the initial venue when the timeout kicks in using the `useReset` hook.
> 
> Why are the changes to the `useReset` hook necessary in order to remove the minZoom when in Wayfinding mode? I seems like we are bundling an unrelated fix into this pull request.

I reverted the changes @ammapspeople. You are right, let's take one bite at the time.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
matbmapspeople: Yes, we need to set minZoom on initial load of the Map Template. Otherwise, we could zoom out until 1.

packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -127,6 +148,12 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped, onRouteFinishe
                 } else {
                     setDestinationDisplayRule(mapsIndoorsInstance.getDisplayRule(directions.destinationLocation));
                 }
+
+                if (mapType === 'mapbox') {
```
ammapspeople, line 152: These lines look very similar to the content of the setMinZoom function. Can we use that?

packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -230,6 +258,28 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped, onRouteFinishe
...
+
+        // Go to current venue once directions are 'Closed'
+        if (isDesktop) {
+            goTo(currentVenue.geometry, mapsIndoorsInstance, 0, 0, getZoomLevel(startZoomLevel), pitch, bearing);
+        } else {
+            getMobilePaddingBottom().then(mobilePaddingBottom => {
+                goTo(currentVenue.geometry, mapsIndoorsInstance, mobilePaddingBottom, 0, getZoomLevel(startZoomLevel), pitch, bearing);
+            });
+        }
+    }
```
ammapspeople, line 270: Are these changes necessary in the context of the issue with minZoom and Wayfinding?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -667,10 +675,27 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+            }
+        } else {
+            getMobilePaddingBottom().then(mobilePaddingBottom => {
+                if (viewModeSwitchVisible) {
+                    goTo(currentVenue.geometry, mapsIndoorsInstance, mobilePaddingBottom, 0, getZoomLevel(startZoomLevel), mapboxPitch, bearing);
+                } else {
+                    goTo(currentVenue.geometry, mapsIndoorsInstance, mobilePaddingBottom, 0, getZoomLevel(startZoomLevel), 0, bearing);
+                }
+            });
+        }
```
ammapspeople, line 698: Are these changes necessary in the context of the issue with minZoom and Wayfinding?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
andreeaceachir142: Why is the value for this 14?

matbmapspeople: Fixed ✅ 
By accidence I flipped 10 with 14.

packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -232,6 +238,20 @@ function Directions({ isOpen, onBack, onSetSize, snapPointSwiped, onRouteFinishe
         setSize(snapPoints.FIT);
     }
 
+    /**
+     * Sets minZoom for a specific map provider.
+     * 
+     * @param {string} mapType 
+     * @param {number} zoomLevel 
+     */
+    function setMinZoom(mapType, zoomLevel) {
```
ammapspeople, line 247: Is the mapType parameter needed? As far as I can tell, there's already a mapType recoil value in the component.

ammapspeople:
Please update the pull request description to reflect the current changes: there is now nothing done in the reset hook or useMapBoundsDeterminer hook.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -239,7 +239,8 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
     const handleMapOptionsChange = (newMapOptions) => {
         setMapOptions(previousMapOptions => ({
             ...previousMapOptions,
-            ...newMapOptions
+            ...newMapOptions,
+            minZoom: 10
```
ammapspeople, line 243: The value 10 for minZoom is used twice. Can we extract it into some constant (file)?

ammapspeople (approved):
✅

PR #487, opened 2025-03-10
Feature: Implement useAppTitle prop

matbmapspeople:
How: 
- Added useAppTitle prop to all the files necessary, as usual. 
- If `useAppTitle && appConfig` are true we set document.title to App Title, otherwise fallback to default. 

Working solution: 

https://github.com/user-attachments/assets/3d3f0787-42cf-4fde-97e1-584988abb1a4

### DIFF
--- a/packages/map-template/CHANGELOG.md
+++ b/packages/map-template/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- New `useAppTitle` prop that Specifices if the app should use title defined by appConfig object.
+- New `useAppTitle` prop that specifies if the Map Template should set the document title as defined in the App Config.
 
 ## [1.71.2] - 2025-03-10
 
--- a/packages/map-template/CHANGELOG.md
+++ b/packages/map-template/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.72.0] - 2025-03-10
+
+### Added
+
+- New `useAppTitle` prop that Specifices if the app should use title defined by appConfig object.
+
 ## [1.71.2] - 2025-03-10
 
 ### Fixed
--- a/packages/map-template/CHANGELOG.md
+++ b/packages/map-template/CHANGELOG.md
@@ -15,8 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- New minZoom vlaue which is 14.
-- Fixed an issue when wayfinding would not show the whole route, but it's part.
+- New minZoom value which is 10.
+- Fixed an issue when wayfinding would not show the whole route.
 
 ## [1.72.3] - 2025-03-13
 
--- a/packages/map-template/index.html
+++ b/packages/map-template/index.html
@@ -5,7 +5,6 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <meta name="theme-color" content="#005655" />
-  <title>MapsIndoors Web</title>
 </head>
 
 <body>
--- a/packages/map-template/index.html
+++ b/packages/map-template/index.html
@@ -5,6 +5,7 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <meta name="theme-color" content="#005655" />
+  <title>MapsIndoors Web</title>
 </head>
 
 <body>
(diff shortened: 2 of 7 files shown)


packages/map-template/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.72.0] - 2025-03-10
+
+### Added
+
+- New `useAppTitle` prop that Specifices if the app should use title defined by appConfig object.
```
ammapspeople, line 12:
```suggestion
- New `useAppTitle` prop that specifies if the Map Template should set the document title as defined in the App Config.
```

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -629,6 +631,13 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setCenter(center);
     }, [center]);
 
+    /**
+     * Sets app title.
+     */
+    useEffect(() => {
+        document.title = (useAppTitle && appConfig) ? appConfig?.appSettings?.title : 'MapsIndoors Web'
```
ammapspeople, line 638: This will always write something to the document title. We agreed that we should not do that.

I would imagine some customers would be surprised to see that their app using the Map Template web component suddenly has the title of "MapsIndoors Web".

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -130,8 +131,9 @@ MapTemplate.propTypes = {
  * @param {boolean} [props.showExternalIDs] - Determine whether the location details on the map should have an external ID visible. The default value is set to false.
  * @param {boolean} [props.searchExternalLocations] - If you want to perform search for external locations in the Wayfinding mode. If set to true, Mapbox/Google places will be displayed depending on the Map Provider you are using. If set to false, the results returned will only be MapsIndoors results. The default is true.
  * @param {string} [props.center] - Specifies the coordinates where the map should load, represented as latitude and longitude values separated by a comma. If the specified coordinates intersect with a Venue, that Venue will be set as the current Venue.
+ * @param {boolean} [props.useAppTitle] - Specifies if the Map Template should set the document title as defined in the App Config.
```
ammapspeople, line 134: Write about the default value.

packages/map-template/index.html
```diff
@@ -5,7 +5,6 @@
   <meta charset="utf-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <meta name="theme-color" content="#005655" />
-  <title>MapsIndoors Web</title>
```
ammapspeople, line 8: Why was this removed?

packages/map-template/src/components/MapsIndoorsMap/MapsIndoorsMap.jsx
```diff
@@ -72,6 +73,7 @@ MapsIndoorsMap.propTypes = {
  * @param {boolean} [props.showRoadNames] - A boolean parameter that dictates whether Mapbox road names should be shown. By default, Mapbox road names are hidden when MapsIndoors data is shown. It is dictated by `mi-transition-level` which default value is 17.
  * @param {boolean} [props.searchExternalLocations] - If you want to perform search for external results in the Wayfinding mode. If set to true, Mapbox/Google places will be displayed depending on the Map Provider you are using. If set to false, the results returned will only be MapsIndoors results. The default is true.
  * @param {string} [props.center] - Specifies the coordinates where the map should load, represented as latitude and longitude values separated by a comma. If the specified coordinates intersect with a Venue, that Venue will be set as the current Venue.
+ * @param {boolean} [props.useAppTitle] - Specifies if the Map Template should set the document title as defined in the App Config.
```
ammapspeople, line 76: Write about the default value.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -629,6 +631,15 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setCenter(center);
     }, [center]);
 
+    /**
+     * Sets app title.
+     */
```
ammapspeople, line 636:
```suggestion
    /*
     * Sets document title based on useAppTitle and appConfig values.
     */
```

It's not a function definition, but a function call. So no JSDoc notation.

Mr.Elite Linter Expert, @Lucaci-Andrei, do you think such a eslint rule exists?

Lucaci-Andrei: I do believe we might have something 🧐

Lucaci-Andrei: Update: unfortunately I could not find a rule that would work only on multiline comments 🤔 . We would need to create a custom rule for it.

PR #488, opened 2025-03-11
Add jest test suite

Lucaci-Andrei:
## What
- Setup the Jest testing framework 
- Add initial test cases

## How
- Add Jest dependencies
- Create test cases for some helper functions, custom hook and a react component


packages/map-template/src/__tests__/components/VenueSelector/VenueSelector.test.jsx
```diff
@@ -0,0 +1,127 @@
...
+        expect(screen.getByText('Main Campus')).toBeInTheDocument();
+        expect(screen.getByText('South Campus')).toBeInTheDocument();
+    });
+
+    it('adds the venue-selector__button--open class when active is true', () => {
+        // Render the VenueSelector with active set to false
+        renderVenueSelector({ active: true });
+
+        // Find the toggle button using its accessible role and name
+        const toggleButton = screen.getByRole('button', { name: /venues/i });
```
ammapspeople, line 94: Clarification:  What does the `name: /venues/i` part do? Does the button have a name containing the string venues or does it mean something else?

Lucaci-Andrei: In this context, the query will look for a button element whose accessible name matches the text /venues/ , the i flag makes sure the match should be case-insensitive

packages/map-template/src/__tests__/helpers/MyPositionLocation.test.js
```diff
@@ -0,0 +1,50 @@
...
+        expect(result.properties).toEqual({
+            name: 'My position',
+            anchor: {
+                type: 'Point',
+                coordinates: [10.19, 56.17]
+            }
+        });
+    });
+
+    it('matches turf point structure', () => {
```
ammapspeople, line 38: I guess it's more about matching a GeoJSON structure 🤔

Lucaci-Andrei: Precisely, we are checking if the generateMyPositionLocation creates the proper object shape 😄

ammapspeople: So shouldn't it be `it('matches GeoJSON Point structure')`?

PR #490, opened 2025-03-13
Fix query param console error

Lucaci-Andrei:
### What
- Map Template would throw errors in the console when using query parameters

### How
- Typecast query parameters to a numerical value to be consistent with the required prop type


ammapspeople (approved):
👏

PR #491, opened 2025-03-14
Fix: startZoomLevel not working for kioskOriginLocationId and locationId

matbmapspeople:
What:
- I found out that we set zoom level to `undefined` inside goTo() method for kiosk and locationId. It is because if we do not set it to undefined, we will be zoomed out too far from a desired location.

How:
- To fix it, I check if startZoomLevel prop is defined for kiosk or locationId. If it is, I let it zoom in to desired startZoomLevel. Otherwise I keep old functionality. 

Working solution:

https://github.com/user-attachments/assets/b27c5e83-3b17-4d15-965b-cb15ad259b81

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21495,7 +21495,7 @@
     },
     "packages/map-template": {
       "name": "@mapsindoors/map-template",
-      "version": "1.72.0",
+      "version": "1.72.3",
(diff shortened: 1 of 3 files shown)


ammapspeople (approved):
✅

PR #492, opened 2025-03-20
Bugfix: Fix pitch and bearing while startZoomLevel is present

matbmapspeople:
There was no calls to SDK to set pitch and bearing when startZoomLevel is present. I added them now.


matbmapspeople:
I could possibly also created a separate function to set pitch and bearing. Let me know what you think 😄

ammapspeople (approved):
✅

PR #493, opened 2025-03-20
Update Sentry DSN configuration to ignore logging for localhost

Lucaci-Andrei:
# What 
- Stop sentry for logging errors in the development environment

# How
- Initialize sentry with the correct dsn if the environment is set to production, otherwise set it to undefined


matbmapspeople (approved):
LGTM ✅ 
Is there any way I can test it?

PR #494, opened 2025-03-21
Sharpen documentation for center prop

ammapspeople:
# What

- Change JSDoc for the center prop
- Sharpen the implementation for converting the center prop to a Point feature

# Why

The JSDoc describing the prop had longitude and latitude in the wrong order. So did the implementation naming-wise, and the JSDoc for that was describing it incorrectly.


andreeaceachir142 (approved):
👏

andreeaceachir142:
@ammapspeople this is ready to be merged in ✅

PR #496, opened 2025-03-25
Bugfix: Fix switching venues on onRouteFinished

matbmapspeople:
What:
As it is right now, `onRouteFinished` we use function that resets everything (useReset.js - used for timeout). This is wrong approach.

How:
I created a new hook, called `useOnRouteFinished.js`. There I perform all necessary states resets when user click onRouteFinished.

Working solution:

Uploading Screen Recording 2025-03-25 at 08.36.25.mov…


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -683,6 +686,10 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         resetState();
         resetAppHistory();
         setResetCount(curr => curr + 1); // will force a re-render of bottom sheet and sidebar.
+    }
+    
+    function onRouteFinish() {
```
ammapspeople, line 691: JSDoc missing.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -720,7 +730,7 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
                         pushAppView={pushAppView}
                         currentAppView={currentAppView}
                         appViews={appStates}
-                        onRouteFinished={() => resetStateAndUI()}
+                        onRouteFinished={() => onRouteFinish()}
```
ammapspeople, line 733: Since we no longer use the resetStateAndUI to clean up then the route is finished, there is maybe no need to call onRouteFinish from the MapTemplate component. There is maybe an opportunity to eliminate some callback spaghetti by just using the hook from the Directions component

ammapspeople (approved):

PR #500, opened 2025-04-02
Feature: Implement normalizeText for mi-dropdown

matbmapspeople:
What:
- It is hard to search for terms like `illy Caffè` using word caffe, because mi-dropdown does not return any results. 

How: 
- Used normalizeText package. 
- It is applied for items in the list and search term so we can find exact matches, without worrying that our keyboard does not support specific symbols/letters.

Testing:
- I added a few test inside components text.html page.

### DIFF
--- a/packages/components/src/components/dropdown/dropdown.tsx
+++ b/packages/components/src/components/dropdown/dropdown.tsx
@@ -501,7 +501,8 @@ export class Dropdown {
                 threshold: -10000
             };
 
-            // Normalize text for search terms
+            // Normalize text by trimming whitespace, converting to lowercase, and removing diacritics.
+            // This ensures consistent matching/comparison of dropdown items.
             const fuzzyResults = fuzzysort.go(normalizeText(inputQuery), miDropdownItemTexts, searchResultsOptions);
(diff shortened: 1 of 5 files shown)


packages/components/src/components/dropdown/dropdown.tsx
```diff
@@ -480,7 +481,8 @@ export class Dropdown {
     filter(): void {
         if (this.filterElement) {
             const inputQuery: string = this.filterElement.value;
-            const miDropdownItemTexts: string[] = this.items.map(item => (item.text || item.innerText));
+            // Normalize text for all the items in the dropdown
```
ammapspeople, line 484: Describe what normalizing means in this context.

packages/components/src/components/dropdown/dropdown.tsx
```diff
@@ -497,8 +500,10 @@ export class Dropdown {
                 allowTypo: false,
                 threshold: -10000
             };
-            const fuzzyResults = fuzzysort.go(inputQuery, miDropdownItemTexts, searchResultsOptions);
-            const filteredItems = fuzzyResults.map(result => this.items.find(item => (item.text || item.innerText) === result.target));
+
+            // Normalize text for search terms
```
ammapspeople, line 504: Consider adding the same explanation about normalizing here.

ammapspeople (approved):
🏅

PR #501, opened 2025-04-08
Fix/add script to respect nvmrc version

Lucaci-Andrei:


.vscode/nvm-init.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+# filepath: /Users/anlu/Projects/Work/web-ui/.vscode/nvm-init.sh
```
ammapspeople, line 2: I guess there is no need to expose how you structure your files on your own machine 😁

Lucaci-Andrei: 🤦

.vscode/nvm-init.sh
```diff
@@ -0,0 +1,29 @@
...
+            nvm use node
+            ;;
+    esac
+fi
+
+# Launch Zsh with the updated PATH
+NODE_BIN="$(nvm which current)"
+export PATH="$(dirname "$NODE_BIN"):$PATH"
+echo "Terminal using $(node -v)"
+exec zsh
```
ammapspeople, line 29: Are we adding another requirement here, that you have to use `zsh`?

Lucaci-Andrei: Yes, at the moment the script works only with zsh. It could be refactored to try to use bash if zsh is not found 🤔 . What do you think?

PR #503, opened 2025-04-22
Fix Sentry version identifiers

Lucaci-Andrei:


packages/map-template/vite.config.js
```diff
@@ -21,11 +23,19 @@ export default defineConfig(() => {
             ViteFaviconsPlugin('./public/favicon.png'),
             eslint(),
             sentryVitePlugin({
-                org: process.env.SENTRY_ORG,
```
ammapspeople, line 24: These secrets exists on the Github repo and are used when building for NPM when releasing. We have to make sure that they are still used correctly.

Lucaci-Andrei: Is the vite config used when building for npm? Or is the buildForNpm script that handles the building?

packages/map-template/.env.example
```diff
@@ -1,2 +1,5 @@
 VITE_MAPBOX_ACCESS_TOKEN=mapbox access token
 VITE_GOOGLE_MAPS_API_KEY=google maps api key
+VITE_SENTRY_ORG=your-org
```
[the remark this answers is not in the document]
Lucaci-Andrei: Added comments that specify that the sentry configurations are optional 👍

packages/map-template/vite.config.js
ammapspeople: Yes to both questions 😁 buildForNpm handles the building using Vite and its config.

Lucaci-Andrei: How could we make sure they are used correctly 🤔?  My guess is that if the sourcemaps are uploaded to sentry, it means that it works ? 🧐

PR #504, opened 2025-04-23
Listen to the app config latitude and longitude values 

Lucaci-Andrei:
## What 
- Two new app config keys were added in the CMS, 'latitude' and 'longitude'
- We need to listen for those values in the map template and react accordingly

## How
- Set up a fallback useEffect that will set the center based on app config's latitude and longitude, if present
- Enhance current useEffect to set center based on map type, since google maps and mapbox order latitude and longitude differently

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -62,7 +62,6 @@ import isNullOrUndefined from '../../helpers/isNullOrUndefined.js';
 import centerState from '../../atoms/centerState.js';
 import PropTypes from 'prop-types';
 import { ZoomLevelValues } from '../../constants/zoomLevelValues.js';
-import mapTypeState from '../../atoms/mapTypeState.js';
 
 // Define the Custom Elements from our components package.
 defineCustomElements();
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -227,8 +227,8 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
 
             const miSdkApiTag = document.createElement('script');
             miSdkApiTag.setAttribute('type', 'text/javascript');
-            miSdkApiTag.setAttribute('src', 'https://app.mapsindoors.com/mapsindoors/js/sdk/4.39.2/mapsindoors-4.39.2.js.gz');
-            miSdkApiTag.setAttribute('integrity', 'sha384-+Lvi/laUwP/QZixVuIXpTn6ckln8itxc4ErCg9xwYY/QSyTqPTgGGqnLkfnhS2zo');
+            miSdkApiTag.setAttribute('src', 'https://app.mapsindoors.com/mapsindoors/js/sdk/4.39.0/mapsindoors-4.39.0.js.gz');
+            miSdkApiTag.setAttribute('integrity', 'sha384-3S9Jvub8zrQ8mn1GlIvVw+LQTsvSF9tz/1e+mr18/rHZfAjUsQ7Vy6cnzidMoyq7');
             miSdkApiTag.setAttribute('crossorigin', 'anonymous');
             document.body.appendChild(miSdkApiTag);
             miSdkApiTag.onload = () => {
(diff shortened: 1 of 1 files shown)


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -628,10 +631,46 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+     * 2. The center prop passed directly to the component
+     * The coordinates are formatted based on the map provider:
+     * - For Google Maps: "latitude,longitude"
+     * - For Mapbox: "longitude,latitude"
+     */
+    useEffect(() => {
+        if (center) {
+            const [latitude, longitude] = center.split(',');
+            let formattedCenter;
+            if (mapType === 'mapbox') {
```
ammapspeople, line 645: I don't think it should be necessary to handle different map providers here. Ultimately, the center prop is passed down all the way to the `useMapBoundsDeterminer` custom hook, which then calls either the SDK's goTo method - with a point geometry, which has latitude and longitude specified - or the mapView's setCenter method, also with latitude and longitude specified.

If we have an ordering bug, I propose we fix it in the useMapBoundsDeterminer hook. I'd like to keep the MapTemplate component as map agnostic as possible.

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -628,10 +631,46 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+     * Fallback mechanism for setting the center position.
+     * If no center is specified (neither through URL parameter nor prop),
+     * and the appSettings contains both latitude and longitude,
+     * use those coordinates as the center position.
+     * The coordinates are formatted based on the map provider:
+     * - For Google Maps: "latitude,longitude"
+     * - For Mapbox: "longitude,latitude"
+     */
+    useEffect(() => {
+        if (!center && appConfig?.appSettings?.latitude && appConfig?.appSettings?.longitude) {
```
ammapspeople, line 664: Would it be simpler if the appConfig value was stored in the same format as the prop? ("represented as longitude and latitude values separated by a comma")

Example: `center: 12.34,56.78`

Lucaci-Andrei: Fixed 👍

ammapspeople: 🥳

ammapspeople: But we still maintain to save latitude and longitude as separate properties in the appConfig instead of storing in in a comma-separated center property?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
Lucaci-Andrei: Fixed 👍 , the fallback useEffect has been simplified as well

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
Lucaci-Andrei: They are saved as individual properties in the appConfig 🤔

ammapspeople: Should they? 😛

Lucaci-Andrei: They should not 😄 . I have updated the branch to listen for a string value and react accordingly 👍

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -633,6 +633,18 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setCenter(center);
     }, [center]);
 
+    /*
+     * Fallback mechanism for setting the center position.
+     * If no center is specified (neither through URL parameter nor prop),
```
ammapspeople, line 638: I don't think we should talk about URL parameters here. Any URL parameter is "transformed" to a prop in MapsIndoorsMap and passed on if the supportsUrlParameters prop is set to true. So talking about them here seems a little "late" 😄

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -633,6 +633,18 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+     * and the appSettings contains both latitude and longitude,
```
ammapspeople, line 639: I believe this is not structured like this anymore, or is it?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -558,7 +558,7 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
     }, [kioskOriginLocationId, mapsindoorsSDKAvailable]);
 
     /*
-     * React on changes to the timout prop
```
ammapspeople, line 561: 😂

PR #505, opened 2025-04-24
Add sentry error boundary for map template

Lucaci-Andrei:
## What
- Improve sentry error logging
## How
-  Add sentry error boundary around our mapsindoors map component
- Add rule to ignore sdk in error logging

### DIFF
--- a/packages/map-template/src/App.jsx
+++ b/packages/map-template/src/App.jsx
@@ -4,13 +4,15 @@ import MapsIndoorsMap from './components/MapsIndoorsMap/MapsIndoorsMap';
 
 function App() {
     return (
-        <div className="app">
-            {/* This is the Map Template component */}
-            <MapsIndoorsMap supportsUrlParameters={true}
-                gmApiKey={import.meta.env.VITE_GOOGLE_MAPS_API_KEY}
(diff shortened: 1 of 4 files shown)


packages/map-template/src/App.jsx
```diff
@@ -4,13 +4,15 @@ import MapsIndoorsMap from './components/MapsIndoorsMap/MapsIndoorsMap';
 
 function App() {
     return (
-        <div className="app">
-            {/* This is the Map Template component */}
-            <MapsIndoorsMap supportsUrlParameters={true}
-                gmApiKey={import.meta.env.VITE_GOOGLE_MAPS_API_KEY}
-                mapboxAccessToken={import.meta.env.VITE_MAPBOX_ACCESS_TOKEN}
+        <Sentry.ErrorBoundary>
```
ammapspeople, line 7: App.jsx is not included when building for NPM. I would imagine that the error boundary should be moved to the MapsIndoorsMap file.

Lucaci-Andrei: Good catch 😄 . It has been moved to MapsIndoorsMap 👍

ammapspeople (approved):
👏

PR #506, opened 2025-04-28
Feature: Nested Categories

matbmapspeople:
UX/UI-wise approved (had a quick check with Oliver, still needs to be properly tested)

**What:** 
It was requested to introduce Nested Categories in the Map Template. So Category can have multiple sub-categories. 

**Requirements:**
- Category can have multiple sub-categories.
- One sub-category can belong to multiple parents.
- Parent category cannot become a someone else child.
- When selecting Category, sub-categories (if any) and Category specific Locations should be displayed. 
- When selecting sub-category, sub-category specific Location should be displayed.
- When selecting Category and sub-category, first 'Go back' click should forward us to parent category and second click should forward us to initial Categories view. 

**How (implementation order):**
- Inside `useCurrentVenue.js` hook, I made a call to the Web SDK to get solutions' specific categories. Then I compared the Web SDK keys with App Config categories that the Map Template uses, and added `childKeys` to all necessary places.
- Inside `Categories.jsx`, I made it so on initial Categories view, those Categories who are someone else child are not visible. They are simply filtered out.
- Inside `Search.jsx`, I firstly worked on what this component returns. I made it so when parent Category is selected, we are shown list of sub-categories and list of Locations that are using this parent Category. Secondly, I added `selectedCategoriesTree` const that keeps track of selected categories. It is an array that stores first clicked category (parent) and second clicked category (child). The purpose of this property is to make sure that we get correct Locations and we can smoothly move between first-second level of Categories. 
- Lastly, I added some styling to display sub-categories and locations correctly.

PS.: I left `MapTemplate.jsx` changes so I can continue my work. It will be cleaned once it is verified that production works. 

**Working solution:**

https://github.com/user-attachments/assets/1c45387d-61ac-4e94-b06e-dc59b5f0398e


packages/map-template/src/hooks/useCurrentVenue.js
```diff
@@ -80,7 +80,18 @@ export const useCurrentVenue = () => {
     /**
      * Generate list of categories that exist on Locations in the current Venue.
```
ammapspeople, line 81: Explain the strategy (and the why) of reading both categories and mainMenu.

packages/map-template/src/hooks/useCurrentVenue.js
```diff
@@ -80,7 +80,18 @@ export const useCurrentVenue = () => {
...
      */
-    const updateCategories = () => {
+    const updateCategories = async () => {
+        const categories = await window?.mapsindoors?.services?.SolutionsService?.getCategories();
+        const mainmenu = appConfig?.menuInfo?.mainmenu ?? [];
```
ammapspeople, line 85:
```suggestion
        const mainMenu = appConfig?.menuInfo?.mainmenu ?? [];
```

Let's not continue the misspelling 😁

packages/map-template/src/hooks/useCurrentVenue.js
```diff
@@ -80,7 +80,18 @@ export const useCurrentVenue = () => {
...
+        const categoriesMap = new Map(
+            categories.map(cat => [cat.key, cat.childKeys])
```
ammapspeople, line 87: Meow 🐱 

Consider not using abbreviations.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -117,23 +117,33 @@ function Search({ onSetSize, isOpen }) {
 
     const [isInputFieldInFocus, setIsInputFieldInFocus] = useState();
 
+    const selectedCategoriesTree = useRef([]);
```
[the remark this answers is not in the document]
matbmapspeople: I applied better wording.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -517,52 +544,72 @@ function Search({ onSetSize, isOpen }) {
...
+                                                    {childInfo.displayName}
+                                                </button>
+                                            </div>
+                                        ))}
+                                </div>
+                            )}
+                        </>
+                    )}
 
+                    {/* Render list of Locations for category/sub-category. */}
```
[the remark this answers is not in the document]
matbmapspeople: Good catch

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -517,52 +544,72 @@ function Search({ onSetSize, isOpen }) {
...
-                            key={location.id}
-                            location={location}
-                            locationClicked={() => onLocationClicked(location)}
-                            isHovered={location?.id === hoveredLocation?.id}
-                        />
-                    )}
-                </div>
-            }
 
+                    {/* If parent category is selected, render back button, sub-categories (if any) and Locations for that parent category */}
```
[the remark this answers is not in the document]
matbmapspeople: A parent category is one that has a non-empty childKeys array. If the childKeys array is empty or undefined, the category is not a parent to any other category. Would you like me to change this comment to make it more understandable? @ammapspeople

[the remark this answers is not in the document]
ammapspeople: In that case, I think just saying "if a category is selected" would suffice.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -491,6 +517,16 @@ function Search({ onSetSize, isOpen }) {
         }
     }, [kioskLocation]);
 
+    /**
+     * 
+     */
+    useEffect(() => {
+        const selectedCategoryInfo = categories.find(([key]) => key === selectedCategory)?.[1];
+        const childKeys = selectedCategoryInfo?.childKeys || [];
```
ammapspeople, line 525: I think these two statements can be merged into one.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -517,52 +553,70 @@ function Search({ onSetSize, isOpen }) {
...
-                            isHovered={location?.id === hoveredLocation?.id}
-                        />
-                    )}
-                </div>
-            }
 
+                    {/* If category is selected, render back button, sub-categories (if any) and Locations for that parent category */}
+                    {selectedCategory && (
+                        <>
+                            <div className="search__nav">
```
ammapspeople, line 577: "search__nav" could be its own BEM Block, "search-nav". Then we can also use proper BEM Elements for the button and text.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -517,52 +553,70 @@ function Search({ onSetSize, isOpen }) {
...
+                                <button aria-label='Back' type="button" className="search__nav-button" onClick={handleBack}>
+                                    <ChevronLeft />
+                                </button>
+                                <div className="search__nav-text">
+                                    {categories?.find(([category]) => category === selectedCategory)[1]?.displayName}
+                                </div>
+                            </div>
+
+                            {/* Show child category only if selectedCategory is NOT a child category to itself */}
+                            <div className='search__results--sub-categories'>
```
ammapspeople, line 587: The class name is written as a BEM modifier, but is a BEM element.

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -517,52 +553,70 @@ function Search({ onSetSize, isOpen }) {
...
-                        />
-                    )}
-                </div>
-            }
 
+                    {/* If category is selected, render back button, sub-categories (if any) and Locations for that parent category */}
+                    {selectedCategory && (
+                        <>
+                            <div className="search__nav">
+                                <button aria-label='Back' type="button" className="search__nav-button" onClick={handleBack}>
```
ammapspeople, line 578: We must use the translation function `t()` for the "Back" text.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -137,22 +141,33 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
...
 
+    /**
+     * Collects all child keys from all categories.
+     * Filters out categories from the initial view that are listed as a child to another category.
+     */
+    useEffect(() => {
+        const childKeys = categories.flatMap(([, category]) => category.childKeys || []);
+        const categoriesWithoutChildKeys = categories.filter(([key]) => !childKeys.includes(key));
+        setCategoriesWithoutChildKeys(categoriesWithoutChildKeys)
+    }, [categories])
```
ammapspeople, line 152:
```suggestion
        setCategoriesWithoutChildKeys(categoriesWithoutChildKeys);
    }, [categories]);
```

packages/map-template/src/components/Search/Search.jsx
```diff
@@ -517,52 +552,55 @@ function Search({ onSetSize, isOpen }) {
...
+                            topLevelCategory={false}
+                            selectedCategoriesArray={selectedCategoriesArray}
                         />
                     )}
-                </div>
-            }
-
 
+                    {/* Show locations when there are any searchResults */}
+                    <div className='search__results--locations'>
```
ammapspeople, line 587: This is notation for a BEM Modifier used as a BEM Element.

ammapspeople (approved):
🆗

PR #508, opened 2025-05-01
Feature/Nested Categories  -- simplified version

matbmapspeople:
What:
- I simplified the code a little bit and now Categories.jsx file is responsible for displaying parent and child categories. We now, pass all necessary information to Category component.
- There is a lot of code that was cleaned up in Search.jsx and Search.scss.
- Behavior-wise all is the same.


matbmapspeople:
@ammapspeople it is ready for review. As we have talked, `Category.jsx` component is responsible for distinguishing between parent/child categories and displaying them correctly. Let me know if you have any questions.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -31,7 +35,7 @@ Categories.propTypes = {
  * @param {object} props.searchFieldRef - The reference to the search input field.
  * @param {boolean} props.isOpen - Determines wheteher the Categories window is open or not.
  */
-function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen }) {
+function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen, topLevelCategory, handleBack, selectedCategoriesArray }) {
```
ammapspeople, line 38: Add them as JSDoc params too.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -151,19 +159,60 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
         setCategoriesWithoutChildKeys(topLevelCategories);
     }, [categories]);
 
+    /**
```
ammapspeople, line 162:
```suggestion
    /*
```

Not a function definition, but a function call.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -151,19 +159,60 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
...
+     * Collects all child keys from all categories.
+     * Filters out categories from the initial view that are not listed as a child to another category.
+     */
+    useEffect(() => {
+        const childKeys = categories.flatMap(([, category]) => category.childKeys || []);
+        const childLevelCategories = categories.filter(([key]) => childKeys.includes(key));
+        setCategoriesWithChildKeys(childLevelCategories);
+    }, [categories]);
+
+    /**
```
ammapspeople, line 172:
```suggestion
    /*
```

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -57,6 +61,10 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
 
     const [categoriesWithoutChildKeys, setCategoriesWithoutChildKeys] = useState([]);
 
+    const [categoriesWithChildKeys, setCategoriesWithChildKeys] = useState([]);
```
ammapspeople, line 64: Is it necessary to have separate states for the categories to show (categoriesWithoutChildKeys, categoriesWithChildKeys) instead of treating them as one state, eg. `categoriesToShow` where the content is decided by the topLevelCategory prop?

This would have the benefits of:
- an opportunity to simplify the JSX by removing some duplicate code
- an opportunity to have only one useEffect on the categories state

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -151,19 +159,60 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
...
                                     {categoryInfo.displayName}
                                 </button>
                             </div>
-                        ))}
+                        ))
+                    ) : (
+                        <>
+                            <div className="categories__nav">
+                                <button
+                                    aria-label="Back"
```
ammapspeople, line 197: We must use the translation function `t()` for textual content that is not coming from MapsIndoors data.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -141,29 +153,61 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
...
-        setCategoriesWithoutChildKeys(topLevelCategories);
-    }, [categories]);
+        const categoriesToDisplay = topLevelCategory ? categories.filter(([key]) => !childKeys.includes(key)) : categories.filter(([key]) => childKeys.includes(key));
+        setCategoriesToShow(categoriesToDisplay)
+    }, [categories])
+
+    /*
+     * Gets display name of selected category.
+     */
+    useEffect(() => {
```
ammapspeople, line 170: This could be moved into the other useEffect with just `categories` as a dependency. No need to have two.

packages/map-template/src/components/Search/Categories/Categories.jsx
```diff
@@ -141,29 +153,61 @@ function Categories({ onSetSize, getFilteredLocations, searchFieldRef, isOpen })
...
+                                    aria-label={t('Back')}
+                                    type="button"
+                                    className="categories__nav--button"
+                                    onClick={handleBack}>
+                                    <ChevronLeft />
+                                </button>
+                                <div>{selectedCategoryDisplayName}</div>
+                            </div>
+                            {selectedCategoriesArray.current.length === 1 &&
+                                categoriesToShow.map(([category, categoryInfo]) => (
```
ammapspeople, line 201: Could we change the JSX so we don't repeat the `{categoriesToShow.map(([category, categoryInfo]) => ( <div /> ))}` part?
I imagine the `<div className="categories__nav" />` could be rendered conditionally based on the topLevelCategory prop.

ammapspeople (approved):
Noice! 👍

PR #509, opened 2025-05-06
Fix issue with gmMapId not being respected on the Map Template

Lucaci-Andrei:
## What 
- gmMapId value was not respected in the map template
## How 
- Add gmMapId prop in the missing components


ammapspeople (approved):
👍

PR #512, opened 2025-05-14
Fix bottom sheet not behaving correctly when exiting categories

Lucaci-Andrei:
## What 
- Bottom sheet "jumps" to max size when exiting location details
## How
- Remove unnecessary auto height style


ammapspeople (approved):
🆗 👌

PR #514, opened 2025-05-15
Listen for center from appconfig

Lucaci-Andrei:
## What
- Listen for the center value coming from the appConfig

## How
- Update the setCenter useEffect to listen for appConfig changes


ammapspeople (approved):
👌

PR #516, opened 2025-05-19
Fix location details scrolls to the top when clicking on Opening hours toggle

Lucaci-Andrei:
…when opening hours list is expanded


matbmapspeople (approved):
LGTM

PR #517, opened 2025-05-22
Feature: Fix Finish Route button ending up in initial venue

matbmapspeople:
What: There is an issue that when you run the Map Template, change the Venue, create a route and click Finish Route - you would end up in initial Venue (they one that loaded initially)

How: I fixed it by creating a hook that handles onRouteFinished and I used it inside MapTemple.jsx. 

Solution is similar to previous one, but it's simplified. It also covers the issues that Andrei found: https://mapspeople.atlassian.net/browse/SCE-437

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -62,6 +62,7 @@ import isNullOrUndefined from '../../helpers/isNullOrUndefined.js';
 import centerState from '../../atoms/centerState.js';
 import PropTypes from 'prop-types';
 import { ZoomLevelValues } from '../../constants/zoomLevelValues.js';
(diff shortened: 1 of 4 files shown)


packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -686,6 +689,16 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setSelectedCategory(null); // unselect category when route is finished
     }
 
+    /**
+     * Function that handles the finishing of route.
+     */
+    function onRouteFinish() {
+        finishRoute();
+        resetAppHistory();
+        setResetCount(curr => curr + 1);
```
ammapspeople, line 698: I think it deserves the same explanation in a comment as in `resetStateAndUI`, is it is a somewhat controversial thing to do.

ammapspeople (approved):
🎯

PR #518, opened 2025-05-22
Reposition kiosk sidebar

Lucaci-Andrei:


packages/components/src/components/scroll-buttons/scroll-buttons.tsx
```diff
@@ -126,7 +126,10 @@ export class ScrollButtons {
                     ref={(el: HTMLButtonElement | null): void => {
                         this.upButtonElement = el as HTMLButtonElement | null;
                     }}
-                    onClick={(): void => this.updateScrollPosition(-this.scrollLength)}>
+                    onClick={(event): void => {
+                        event.stopPropagation();
```
matbmapspeople, line 130: Just a comment/question:
How come was it working before and now it is not, and we have to add this?

matbmapspeople: Discussed in the private conversation about it.

matbmapspeople (approved):

PR #519, opened 2025-05-23
Refactor of the Bottom Sheet

ammapspeople:
A refactor of the `BottomSheet`, `Sheet` and it's underlying components in order to iron out a range of layout bugs.

**The BottomSheet component**

- Cleaned up and removed the need for:
    - holding the individual sheet sizes in the `BottomSheet` component.
    - The `reFitWhenContentChanges` callback prop
    - The `snapPointSwiped` callback prop
- Created imperative method on the sheet for setting a sheet snap point, instead of having a not so reliable prop.
- Replaced `preferredSnapPoint` with `initialSnapPoint`. The preferred one could change. Initial should not, but instead the imperative method can be used.
- Renamed `minHeight` to `minimizedHeight` so signify that it's the height in minimized snap point, not just a more general minimum height

**The Sheet component**

- Refactored `changeSheetHeight` function for (hopefully) more robustness.
- Made an (this is unfortunately necessary) clone of the sheet's child prop in order to pass on the `snapPointSwipedByUser` prop.
- Re-implemented the calculation of snap point when the user makes a swipe gesture.
	- This includes unit tests.
	- Due to us not really having unit tests as part of pipeline, you'll need to modify some stuff in the code to run them. It's all described though.

**The Sidebar component**

- Use more descriptive key naming for the sheets and modals

**Other**

- Refactored the underlying components to accommodate the new signature of the Sheet
- Added a convenient `start:network` NPM script to more easily make the Map Template be exposed on the local network for testing. 
- Various layout issues, especially with Location Details and Search


matbmapspeople:
Is it expected to hide bottom sheer once Search field is clicked?

https://github.com/user-attachments/assets/3195acf5-d163-4f66-b454-2e4aae3f0f8d

packages/map-template/src/components/BottomSheet/Sheet/Sheet.jsx
```diff
@@ -1,259 +1,333 @@
-import { useContext, useEffect } from 'react';
+import React, { cloneElement, forwardRef, useContext, useEffect, useImperativeHandle } from 'react'; // eslint-disable-line
```
matbmapspeople, line 1: Is that comment needed there? I removed it, and nothing was highlighted as a warning/error 🤔

packages/map-template/src/components/BottomSheet/Sheet/Sheet.jsx
```diff
@@ -1,259 +1,333 @@
...
+            {/* We need to have a div in a div for it being able to scroll content separately and for the height to be measurable */}
+            <div ref={contentRef} className={`sheet__content ${snappedTo.current === snapPoints.MIN ? 'sheet__content--no-scroll' : ''}`} style={style}>
+                {clonedChildren}
+            </div>
         </div>
-    </div>
+    );
+});
+
+Sheet.propTypes = {
```
matbmapspeople, line 249: Can we move PropTypes at the beginning on the file? I see this is applied for most of file + it is easier to read and figure out stuff.

packages/map-template/src/components/BottomSheet/Sheet/Sheet.jsx
ammapspeople: Sharp eyes 👀 No, it's not needed. Will remove.

packages/map-template/src/components/BottomSheet/Sheet/Sheet.jsx
ammapspeople: Sadly not when the component is wrapped in a `forwardRef` call 😞 (which we need for the imperative `setSnapPoint` function to work)

ammapspeople:
> Is it expected to hide bottom sheer once Search field is clicked?
> Screen.Recording.2025-05-27.at.08.27.01.mov

In this case, the bottom sheet is fitting to the size of the search field + categories. I'll see if I can somehow preserve the full height in case the user has swiped to full size.

ammapspeople:
> > Is it expected to hide bottom sheer once Search field is clicked?
> > Screen.Recording.2025-05-27.at.08.27.01.mov
> 
> In this case, the bottom sheet is fitting to the size of the search field + categories. I'll see if I can somehow preserve the full height in case the user has swiped to full size.

@matbmapspeople This has now been fixed. We also decided always to maximize the bottom sheet when the search field is active.

matbmapspeople (approved):
I cannot find anything outstanding. Also, everything works as expected when testing.

PR #521, opened 2025-05-28
fix: make location details header sticky

Lucaci-Andrei:


matbmapspeople (approved):
Looks good. Remember to add a `changlog` and run build if needed 😄

PR #526, opened 2025-06-04
Bump non breaking dependencies versions in the components package

Lucaci-Andrei:


packages/components/src/components/my-position/my-position.tsx
```diff
@@ -353,10 +348,10 @@ export class MyPositionComponent {
             return;
         }
 
-        const deviceType: DeviceType = this.parser.getDevice().type;
+        const deviceType: IDevice['type'] = this.parser.getDevice().type;
         this.canBeTracked = (
             typeof window.DeviceOrientationEvent === 'function' &&
-            (deviceType === DeviceType.Phone || deviceType === DeviceType.Tablet) &&
+            (deviceType === 'mobile' || deviceType === 'tablet') &&
```
ammapspeople, line 354: Can we use an enum instead of these strings? 
https://docs.uaparser.dev/api/submodules/enums.html

PR #528, opened 2025-06-06
Bugfix: Prevent slow getRoute calls from overriding faster ones

ammapspeople:
# What

Prevent slower getRoute calls from overtaking faster ones.

# Why

In the case of eg. My Position being used, and the position is far away from the current venue, external directions responses can be extremely slow. This had the result of overriding a faster route call then the slow response finally resolved, leading to a sudden jump to an unexpected route.

# How

- Introduced an `isActive` flag to prevent older ones from overtaking. The cleanup function where `isActive` is set to false is always run before the effect runs again.


matbmapspeople (approved):
Looks good to me ✅

PR #534, opened 2025-06-16
Improve view selector based on feedback

Lucaci-Andrei:


packages/react-components/src/components/MIMap/MapControls/MapControls.jsx
```diff
@@ -104,14 +104,15 @@ function MapControls({ mapType, mapsIndoorsInstance, mapInstance, onPositionCont
...
+                    {viewSelectorPortal}
                     {viewModeSwitchPortal}
                     {myPositionPortal}
                     {floorSelectorPortal}
                 </div>
 
                 {/* Bottom right desktop controls */}
                 <div className="map-controls-container desktop bottom-right">
-                    {viewSelectorPortal}
+                    {/* */}
```
matbmapspeople, line 115: Should this be removed?

Lucaci-Andrei: Yup 👍

matbmapspeople:
> Suggestion: Apply pointer-event: none when in wayfinding. Button looks like it is not disabled @Lucaci-Andrei



> Suggestion: Apply pointer-event: none when in wayfinding. Button looks like it is not disabled @Lucaci-Andrei

Also when I hover the button it's changes background color when focused - making me think that it works 😕

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -768,6 +768,7 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
             mapOptions={mapOptions}
             onMapOptionsChange={handleMapOptionsChange}
             gmMapId={gmMapId}
+            isWayfindingActive={currentAppView === appStates.WAYFINDING || currentAppView === appStates.DIRECTIONS}
```
matbmapspeople, line 771: Should we consider calling this prop differently? 
isWayfindingOrDirectionsActive - or something like that 🤔 
isNavigationActive maybe? 
Or should we leave it as it is?

Lucaci-Andrei: isWayfindingOrDirections maybe 🤔 ?

matbmapspeople: Let's do it 👍

packages/react-components/src/components/MIMap/MapControls/MapControls.jsx
```diff
@@ -104,14 +104,14 @@ function MapControls({ mapType, mapsIndoorsInstance, mapInstance, onPositionCont
...
                 <div className="map-controls-container desktop top-right">
                     {venueSelectorPortal}
+                    {viewSelectorPortal}
                     {viewModeSwitchPortal}
                     {myPositionPortal}
                     {floorSelectorPortal}
                 </div>
 
                 {/* Bottom right desktop controls */}
                 <div className="map-controls-container desktop bottom-right">
```
matbmapspeople, line 114: Should this div and scss class be removed?

Lucaci-Andrei: No, that is where the VIewSelector is being rendered 😄

PR #552, opened 2025-07-30
Fix removing kioskOriginlocationId display rule not reverting at runtime

Lucaci-Andrei:
## What:
- Location display rules remained modified after kioskOriginLocationId was removed at runtime

## How:
- Added cleanup mechanism to track and restore original display rules when kioskOriginLocationId changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where display rules for kiosk locations did not revert correctly after kiosk mode was deactivated.


packages/map-template/src/hooks/useMapBoundsDeterminer.js
```diff
@@ -57,6 +57,8 @@ const useMapBoundsDeterminer = () => {
     const [currentVenueName, setCurrentVenueName] = useRecoilState(currentVenueNameState);
     const isMapReady = useRecoilState(isMapReadyState);
     const [center,] = useRecoilState(centerState);
+    // Store kiosk location info for cleanup
+    const [kioskOriginLocationInfo, setKioskOriginLocationInfo] = useState({ id: null, originalRule: null });
```
ammapspeople, line 61: I think that the naming somehow should suggest that this is about storing a Display Rule.

Also, is the `id` necessary given that we have the Atom for `kioskOriginLocationId`?

Lucaci-Andrei: We are storing the original display rule for that specific id, so we know on which location we should revert the display rule if the prop is removed at runtime 🤔

Lucaci-Andrei: Once we remove the parameter at runtime, we do not really know on which location ID we would need to reset the display rule 🤔

ammapspeople: Yeah, facepalm by me.

ammapspeople (approved):
👍

PR #558, opened 2025-08-14
Add lazy load when loading Map Components

Lucaci-Andrei:
## What
- Added code splitting for MapboxMap and GoogleMapsMap components to reduce initial bundle size
- Implemented React lazy() method to load map provider dependencies only when needed

## How
- Converted MapboxMap and GoogleMapsMap imports to use React's lazy() dynamic import function
- Wrapped lazy components with Suspense boundaries (without fallback for simplicity)
- Maintained portal compatibility by keeping MapControls rendering containers before lazy components load
- Bundle now splits into separate chunks one for the Mapbox chunk and one for the Google Maps chunk
- Users only download the map provider code they actually use, reducing initial load time

## Summary by CodeRabbit

- New Features
  - None.

- Refactor
  - Lazy-load map providers to improve initial page load and defer heavy assets.
  - Remove a global map CSS import from one provider to avoid loading unnecessary styles.

- Bug Fixes
  - Make the view-mode switch resilient when its portal target is missing to prevent render errors.

- Documentation
  - Add changelog entry for version 1.82.0 noting lazy-loading of map providers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23381,7 +23381,7 @@
     },
     "packages/map-template": {
       "name": "@mapsindoors/map-template",
-      "version": "1.81.5",
+      "version": "1.81.6",
       "devDependencies": {
         "@googlemaps/js-api-loader": "^1.15.1",
         "@mapsindoors/components": "*",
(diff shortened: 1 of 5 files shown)


matbmapspeople (approved):
LGTM. Seems like it also work on the map.

PR #562, opened 2025-08-27
Add custom location provider and examples

Lucaci-Andrei:
## Summary by CodeRabbit

* **New Features**
  * mi-my-position now accepts a pluggable customPositionProvider (modern event-based or legacy callback) and exposes setPosition(position) for programmatic updates; improved availability/permission handling and immediate-known-position behavior.

* **New Utilities**
  * Added reusable CustomPositionProvider examples (TS/JS) with validation, event on/off, setPosition and error emission.

* **Platform Integration**
  * Added devicePosition prop across map components and controls to drive provider updates and trigger position update/watch.

* **Documentation**
  * Docs updated with provider interface, options, examples, public APIs and events.


matbmapspeople:
Is it correct that position icon is smaller in the Web App, compared to Web SDK? Just curious :D

Screenshot 2025-08-29 at 09 30 43

packages/components/src/types/position-provider.interface.ts
```diff
@@ -0,0 +1,45 @@
...
+    };
+}
+
+/**
+ * Interface for position providers that can supply device location data.
+ * Supports both legacy callback-style and modern event-style providers.
+ */
+export interface IPositionProvider {
+    // Legacy callback-style interface (for backward compatibility)
+    isAvailable?(): boolean;
```
matbmapspeople, line 25: At first, I was not exactly sure what is available. Could we call it isPositionAvailable?

packages/components/src/types/position-provider.interface.ts
```diff
@@ -0,0 +1,45 @@
...
+    isAlreadyGranted?(): Promise<boolean>;
```
matbmapspeople, line 26: I am not sure what is granted. Do you mean position? If so, what is the difference between position available and granted? This need explanation.

packages/components/src/types/position-provider.interface.ts
```diff
@@ -0,0 +1,45 @@
...
+    listenForPosition?(
+        maxAccuracy: number,
+        positionError: (error?: GeolocationPositionError) => void,
+        positionInaccurate: (accuracy: number) => void,
+        positionRequesting: () => void,
+        positionReceived: (position: GeolocationPosition) => void
+    ): void;
+    stopListeningForPosition?(): void;
+
+    // Modern event-style interface (matches documentation)
```
matbmapspeople, line 36: Wherever documentation is mentioned in the comment, can we get link here? 
It might be that someone in the future will get here, but not in the file where the link is.

packages/components/src/components/my-position/CustomPositionProvider.ts
```diff
@@ -0,0 +1,194 @@
...
+     * Creates an instance of the CustomPositionProvider.
+     *
+     * @param {PositionProviderOptions} options - The options for the position provider.
+     */
+    constructor(options?: PositionProviderOptions) {
+        // Merge user options with default options
+        const {
+            positionMarkerStyles,
+            accuracyCircleStyles,
+            ...rest
```
matbmapspeople, line 66: Can we change this `rest` name to something more meaningful?

packages/components/src/components/my-position/CustomPositionProvider.ts
```diff
@@ -0,0 +1,194 @@
...
+        this._currentPosition = position;
+
+        // Emit position_received event
+        const callbacks = this._listeners.get('position_received') || [];
+        callbacks.forEach(callback => {
+            callback.call(null, { position });
+        });
+    }
+
+    /* Alternative implementation for handling simplified position objects:
```
matbmapspeople, line 156: Is it here because users will see it? Shouldn't this be in documentation as alternative example? What if someone will not reach that point to find it, but would be interested in it from f. ex.: documentation?

packages/components/src/components/my-position/my-position.tsx
```diff
@@ -95,6 +108,18 @@ export class MyPositionComponent {
...
 
+    /**
+     * Sets a custom position. Works with any provider that implements setPosition.
+     */
+    @Method()
+    public async setPosition(position: GeolocationPosition): Promise<void> {
+        if (this.positionProvider?.setPosition) {
+            this.positionProvider.setPosition(position);
+        } else {
+            // Current position provider does not support manual position setting
```
matbmapspeople, line 119: Should we return a warning/error? Otherwise can we remove else?

packages/react-components/src/utils/CustomPositionProvider.js
```diff
@@ -0,0 +1,143 @@
+/**
+ * Default options for the CustomPositionProvider, matching the documentation.
+ */
+const DEFAULT_OPTIONS = {
```
matbmapspeople, line 4: Did you get those default styling values from Oliver or?

packages/react-components/src/utils/CustomPositionProvider.js
```diff
@@ -0,0 +1,143 @@
...
+class CustomPositionProvider {
+    constructor(options = {}) {
+        this._currentPosition = null;
+        this._listeners = new Map();
+
+        // Merge user options with default options
+        const {
+            positionMarkerStyles,
+            accuracyCircleStyles,
+            ...rest
```
matbmapspeople, line 32: The same as before, can we use more meaningful name than `rest`?

matbmapspeople (commented):
First round done @Lucaci-Andrei

packages/react-components/src/utils/CustomPositionProvider.js
Lucaci-Andrei: https://docs.mapsindoors.com/~/changes/452/sdks-and-frameworks/web/other-guides/user-positioning/custom-positionprovider#required-properties

packages/components/src/types/position-provider.interface.ts
Lucaci-Andrei: isAvailable is used by the legacy Geolocation API to check for device position, which is using the same interface . If we would change it we would have to change it in the GeoLocationProvider as well.

packages/components/src/types/position-provider.interface.ts
Lucaci-Andrei: That checks if the user has already granted permission to access the device's location.  Legacy Geolocation provider is using it.

packages/components/src/types/position-provider.interface.ts
matbmapspeople: Alright, no change then if it is like that since

matbmapspeople (approved):
LGTM. It would be nice to have someone test it properly besides us.

Lucaci-Andrei:
> LGTM. It would be nice to have someone test it properly besides us.

I agree 😄

PR #565, opened 2025-09-05
Refactor search component by splitting relevant functionality into standalone components 

Lucaci-Andrei:


matbmapspeople:
> > Seems like there is a lot of code shuffling, mostly. So far it looks alright to me, but we need to next it properly. Remember that we have setting Mapbox base map style on map-latest right now, and I do not know if we should merge this in, or not.
> 
> I think we should have that released first then we can promote this one to latest. This refactoring would need some extensive testing

🤝

PR #677, opened 2026-04-10
Fix/qr code in directions

Lucaci-Andrei:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed desktop kiosk QR dialog overlay remaining visible after directions close or kiosk-mode route data is cleared.

### DIFF
--- a/packages/map-template/src/components/Directions/Directions.jsx
+++ b/packages/map-template/src/components/Directions/Directions.jsx
@@ -204,11 +204,22 @@ function Directions({ isOpen, onBack, onSetSize, onRouteFinished }) {
      * Make sure directions stop rendering on the map when the Directions view is not active anymore.
      */
     useEffect(() => {
-        if (!isOpen && directionsRenderer) {
-            stopRendering();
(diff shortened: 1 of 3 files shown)


packages/map-template/src/components/Directions/Directions.jsx
```diff
@@ -204,11 +204,22 @@ function Directions({ isOpen, onBack, onSetSize, onRouteFinished }) {
...
+            }
+        }
+    }, [isOpen, appConfig, setQRCodeLink, isKioskContext]);
+
+    useEffect(() => {
+        if (isKioskContext && !directions) {
+            setQRCodeLink(null);
         }
-    }, [isOpen, appConfig]);
+    }, [directions, isKioskContext, setQRCodeLink]);
```
matbmapspeople, line 222: Should isOpen be here? It's removed

Lucaci-Andrei: Should not be neccesary there

matbmapspeople (approved):

PR #688, opened 2026-05-06
fix: highlight externalIDs locations on initial map load

vishak-mapspeople:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where externalIDs (dot badge) highlights were not displayed on the initial map load. Highlights now wait until the map is fully ready and are re-applied after map transitions, preventing them from being wiped during initial fit/zoom operations and ensuring badges remain visible and correctly positioned during map interactions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MapsPeople/web-ui/pull/688?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/map-template/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mapsindoors/map-template",
   "description": "Get a MapsIndoors map up and running in less than 10 mins.",
-  "version": "1.97.11",
+  "version": "1.97.12",
```
Lucaci-Andrei: it's not necessary to bump versions in package.json, the version bump happens on the CI pipeline side

packages/map-template/src/components/MapWrapper/MapWrapper.jsx
```diff
@@ -203,12 +209,28 @@ function MapWrapper({ onLocationClick, onMapPositionKnown, useMapProviderModule,
...
+                removeIdleListener = () => map.off?.('idle', applyFilterOrHighlight);
+            } else if (mapType === mapTypes.Google && typeof map.addListener === 'function') {
+                const idleListener = map.addListener('idle', () => {
+                    idleListener.remove();
+                    applyFilterOrHighlight();
+                });
+                removeIdleListener = () => idleListener.remove();
+            }
+        }
+
```
Lucaci-Andrei, line 223: On Google Maps, we have a  double-remove. The idle callback self-removes and removeIdleListener removes it again in cleanup. Might be worth dropping the self-remove inside the callback and let cleanup own it exclusively (consistent with how Mapbox works).
  We have a double SDK call for applyFilterOrHighlight. It fires immediately (line 206) and again on idle, so every effect trigger issues two highlight()/filter() calls.
  Asymmetric lifetime management — Mapbox uses map.once, Google uses a manual self-removing wrapper. Might be worth making it consistent.

vishak-mapspeople: Good catches — all three addressed in the latest push.

Self-remove on Google: dropped. The Google listener is now registered as map.addListener('idle', applyFilterOrHighlight) with no inner remove(). Cleanup is the sole owner.
Double SDK call: dropped the immediate applyFilterOrHighlight(). The effect now relies solely on the idle path, so each trigger issues one highlight()/filter() call instead of two.
Asymmetric lifetime: Mapbox switched from map.once to map.on. Both branches now have the same shape — register listener, return remover — and cleanup owns the lifetime exclusively on both providers.

packages/map-template/package.json
vishak-mapspeople: Thanks for letting me know. Reverted in latest push

Lucaci-Andrei (approved):

PR #708, opened 2026-06-19
fix: directionsFrom now selects the origin's venue/building before it's floor

vishak-mapspeople:
Selects the `directionsFrom` origin's venue and building before setting its floor, so the map and the floor selector reliably land on the origin's floor.
  Fixes origins that live in a non-default venue/building, where the selector previously showed the wrong building's floors (e.g. the default venue's ground
  floor).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the directions-from origin location did not reliably sync to the correct floor, venue, and building on the map. The map and floor selector now consistently land on the origin's intended floor.

### DIFF
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -302,7 +302,10 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
             mapsIndoorsInstance.addListener('building_changed', onChange);
             mapsIndoorsInstance.addListener('floor_changed', onChange);
             apply();
-        })();
+        })().catch(error => {
+            detach();
+            console.warn('directionsFrom: failed to sync origin venue/building/floor', error);
+        });
 
         return () => { cancelled = true; detach(); };
     }, [mapsIndoorsInstance, wayfindingOriginHighlightLocation]);
--- a/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
+++ b/packages/map-template/src/components/MapTemplate/MapTemplate.jsx
@@ -255,6 +255,59 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
     }, [directionsFromLocation, directionsTo, directionsFrom]);
 
     /**
+     * directionsFrom origin: drive the SDK to the origin's Venue → Building → Floor.
+     */
+    useEffect(() => {
+        const loc = wayfindingOriginHighlightLocation;
+        if (!mapsIndoorsInstance || !loc) return;
+        const targetFloor = loc.properties.floor;
+        let cancelled = false;
+        let detach = () => {};
+
+        (async () => {
+            const services = window.mapsindoors.services;
+            const venues = await services.VenuesService.getVenues();
+            if (cancelled) return;
+            const venue = venues.find(v =>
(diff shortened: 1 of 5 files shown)


vishak-mapspeople:
@Lucaci-Andrei 

Two known issues intentionally left out of this PR — flagging for follow-up, not fixed here:

  1. **Red pin shows on all floors.** The `directionsFrom` origin pin (`selectLocation`) stays visible on every floor, not just the origin's — kept as-is for now.
  Is that the intended behaviour for an origin? Noting it because kiosk's pin (a display-rule marker) *is* floor-specific, so the two are inconsistent. The same persistent-pin behaviour also applies to the `locationId` path (it uses`selectLocation` as well).

  2. **Same wrong-floor-on-load issue exists for `locationId` and `kioskOriginLocationId`.** Those URL params use the same "set the floor before the building is current" pattern that this PR fixes for `directionsFrom`, so they can also land on the wrong floor when the target is in a non-default venue/building. Not fixed here to keep this PR scoped — good candidate for a shared helper in a separate PR.
  
  Thoughts on this?

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -254,6 +254,59 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
...
+            };
+
+            detach = () => {
+                mapsIndoorsInstance.removeListener('building_changed', onChange);
+                mapsIndoorsInstance.removeListener('floor_changed', onChange);
+            };
+            mapsIndoorsInstance.addListener('building_changed', onChange);
+            mapsIndoorsInstance.addListener('floor_changed', onChange);
+            apply();
+        })();
```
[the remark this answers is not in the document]
vishak-mapspeople: Good catch. Addressed!

packages/map-template/src/components/MapTemplate/MapTemplate.jsx
```diff
@@ -254,6 +254,62 @@ function MapTemplate({ apiKey, gmApiKey, mapboxAccessToken, venue, locationId, p
         setWayfindingOriginHighlightLocation(eligible ? directionsFromLocation : null);
     }, [directionsFromLocation, directionsTo, directionsFrom]);
 
+    /**
+     * directionsFrom origin: drive the SDK to the origin's Venue → Building → Floor.
+     */
+    useEffect(() => {
```
Lucaci-Andrei: Could we create a custom hook to use it here and improve the scoped variable names ?

vishak-mapspeople: @Lucaci-Andrei  Good call — happy to extract this into a custom hook and tidy up the scoped variable names.

One scope question before I do: should I keep it scoped to `directionsFrom` for now, or shape it as the shared helper I flagged in my earlier comment, so `locationId` and `kioskOriginLocationId` can reuse it for the same wrong-floor-on-load fix?

Lucaci-Andrei: If we can centralize it better so they share the same helper I say go for it 😄

vishak-mapspeople: @Lucaci-Andrei 
Refactored into custom hook and committed

Lucaci-Andrei (approved):

PR #710, opened 2026-06-24
Bugfix/fix setting context analytics

matbmapspeople:


packages/map-template/src/helpers/syncAnalyticsViewVariant.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Resolve the analytics viewVariant from host props.
+ * Uses kioskOriginLocationId (sync) rather than the fetched kiosk Location object.
+ *
+ * @param {string | undefined} kioskOriginLocationId
+ * @param {boolean} isDesktop
+ * @returns {'kiosk' | 'desktop' | 'mobile'}
+ */
+export function getAnalyticsViewVariant(kioskOriginLocationId, isDesktop) {
+    if (kioskOriginLocationId && isDesktop) {
```
Lucaci-Andrei, line 10: might be worth checking if the id is valid (i.e location exists in our data) before tagging to avoid any potential false positives

packages/map-template/src/helpers/syncAnalyticsViewVariant.js
```diff
@@ -0,0 +1,31 @@
...
+        return 'kiosk';
+    }
+
+    return isDesktop ? 'desktop' : 'mobile';
```
Lucaci-Andrei, line 14: similar approach here,  worth double checking when that hook fires during initialization to be sure the right view variant is targeted

Lucaci-Andrei (approved):

PR #711, opened 2026-06-25
Feat/improve accesibility v2

Lucaci-Andrei:


packages/map-template/src/components/ChatWindow/ChatMessages/ChatMessages.scss
```diff
@@ -25,7 +25,7 @@
     padding: var(--spacing-x-small) var(--spacing-medium);
     align-self: flex-end;
     background-color: var(--chat-window-primary-color, #e6f7ff);
-    color: var(--color-white-white);
+    color: var(--chat-window-primary-text-color, #1a1a1a);
```
vishak-mapspeople, line 28: Contrast regression (WCAG 1.4.3, AA). --chat-window-primary-text-color is never set anywhere, so this always falls back to dark #1a1a1a. The bubble background is the brand primaryColor (default dark teal #005655) — so on the default theme this is dark text on a dark bubble. The previous --color-white-white was correct here. Verified live.

packages/map-template/src/components/Accessibility/Accessibility.a11y.test.jsx
```diff
@@ -0,0 +1,19 @@
+import { render } from '@testing-library/react';
+import { toHaveNoViolations } from 'jest-axe';
+
```
vishak-mapspeople, line 5: This test doesn't exercise the component. It renders hand-written stub JSX and asserts on that, so it passes regardless of whether <ComponentName> is correct — false confidence. Not sure if this is intended😅

Applicable to : Accessibility.a11y.test.jsx, ChatInput.a11y.test.jsx, LegendDialog.a11y.test.jsx, LocationDetails.a11y.test.jsx, MapTemplate.a11y.test.jsx, Search.a11y.test.jsx, MapTemplate.lang.test.jsx, hooks/useInactive.test.js, hooks/useReturnFocus.test.js.
(QRCodeDialog.a11y.test.jsx is the one that does it right)

packages/map-template/src/components/Wayfinding/Wayfinding.jsx
```diff
@@ -613,10 +614,10 @@ function Wayfinding({ onStartDirections, onBack, directionsToLocation, direction
                             changed={() => onInputChanged(searchFieldIdentifiers.TO)}
                         />
                     </label>
-                </div>
+                </fieldset>
             </div>
-            {!hasFoundRoute && <p className="wayfinding__error">{t('No route found')}</p>}
-            {!hasSearchResults && !showMyPositionOption && <p className="wayfinding__error">{t('Nothing was found')}</p>}
+            {!hasFoundRoute && <p className="wayfinding__error" role="status" aria-live="polite">{t('No route found')}</p>}
```
vishak-mapspeople, line 619: Focus not returned on close (WCAG 2.4.3, A). Verified live: closing Wayfinding leaves document.activeElement = <body>. useReturnFocus was added to Directions/LocationDetails/Chat but not here. Add it (same pattern)

```
// with the imports:
 import { useReturnFocus } from '../../hooks/useReturnFocus';

/ first line inside the component, after `const { t } = useTranslation();`:
useReturnFocus(isActive);
```

vishak-mapspeople (commented):
- Chat-bubble contrast regression (1.4.3) — dark-on-dark on the default theme. 
- New a11y tests assert on stubs, not components — false confidence.

**Real gaps (in scope)**

- Wayfinding doesn't return focus on close (2.4.3) — inline on Wayfinding.jsx. (Check Search/VenueSelector/LanguageSelector too.)
- Map has no visible focus indicator (2.4.7). The Mapbox canvas is keyboard-pannable, but no focus ring. Css fix should be fine
- lang/title clobber host pages (3.1.1/2.4.2) — gate behind a prop. Inline on MapTemplate.jsx.

**Out of scope**

- Reduced-motion ignores map animations (2.3.3) — fly-to/route-draw live in the MapsIndoors JS SDK, not here. SDK ticket.
- Search results not keyboard-operable (2.1.1) — mi-list-item-location in @mapsindoors/components (no tabindex/keydown). Components-package ticket.

 **Verified fine (credit due)**

- Map keyboard pan works on Mapbox at runtime (2.1.1 ✓). Target sizes are 48px / ≥24px (2.5.8 ✓).
- Solid: alt text, headings, fieldset/labels, aria-live regions, dialog focus traps, reflow/zoom/text-spacing, scroll-padding (2.4.11), and the inactivity warning (what actually makes 2.2.1 pass).

packages/map-template/src/components/Accessibility/Accessibility.a11y.test.jsx
Lucaci-Andrei: Stubs should be fixed now 👍

PR #712, opened 2026-06-30
Add chat reset functionality and warnings when reaching token cap

Lucaci-Andrei:
## Summary by CodeRabbit

* **New Features**
  * Added session token-limit lifecycle states, including “approaching” and “exhausted”.
  * Show a mobile token-limit warning panel with a **Start new session** action.
  * Introduced independent chat input locking via an **input disabled** mode (focus and send controls).
* **Bug Fixes**
  * Session reset clears chat history.
  * When limits are reached, the UI displays the correct translated session-limit message.
* **Documentation / Localization**
  * Updated session-limit text across all supported languages.

### DIFF
--- a/packages/map-template/src/providers/GeminiProvider.jsx
+++ b/packages/map-template/src/providers/GeminiProvider.jsx
@@ -324,7 +324,7 @@ export function GeminiProvider({ children, enabled }) {
         } catch (error) {
             if (error?.message === 'SESSION_TOKEN_LIMIT_EXCEEDED') {
                 setSessionStatus('exhausted');
-                const limitMessage = 'Your session has reached its token limit. Please start a new session to continue.';
+                const limitMessage = 'SESSION_LIMIT_MESSAGE';
                 if (options.onComplete) {
                     options.onComplete({ response: limitMessage, error });
                 }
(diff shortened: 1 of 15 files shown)


packages/map-template/src/providers/GeminiProvider.jsx
```diff
@@ -295,11 +322,21 @@ export function GeminiProvider({ children, enabled }) {
 
             return await handleNonStreamingResponse(sessionId, prompt, extra, onComplete);
         } catch (error) {
+            if (error?.message === 'SESSION_TOKEN_LIMIT_EXCEEDED') {
+                setSessionStatus('exhausted');
+                const limitMessage = 'Your session has reached its token limit. Please start a new session to continue.';
```
vishak-mapspeople: this limit message is hardcoded English, so it shows up as an English chat bubble even in DE/DA/zh UIs (the banner above it is translated). Since you're already touching all 9 locale files, worth adding a key and passing the translated string down

vishak-mapspeople (approved):
Just the translation, rest LGTM ✅

packages/map-template/src/providers/GeminiProvider.jsx
Lucaci-Andrei: Done 😄

PR #715, opened 2026-07-13
Add missing aria-labels, update translations to missing keys

Lucaci-Andrei:
## Summary by CodeRabbit

* **Accessibility**
  * Improved screen-reader support across chat, navigation, map view, language, QR code, legend, and kiosk controls.
  * Added announcements for dynamic content, errors, route steps, search results, and AI responses.
  * Added clearer dialog semantics and state indicators for toggles and expandable panels.

* **Localization**
  * Added translated accessibility labels and interface text across nine supported languages.

### DIFF
--- a/packages/map-template/src/components/ChatButton/ChatButton.jsx
+++ b/packages/map-template/src/components/ChatButton/ChatButton.jsx
@@ -47,7 +47,7 @@ function ChatButton({ pushAppView, currentAppView, appViews }) {
         <button 
             className="chat-button"
             onClick={handleOpenChat}
-            title="Open chat"
+            title={i18n.t('Open chat')}
             aria-label={i18n.t('Open chat')}
             type="button"
             style={{ '--chat-button-primary-color': primaryColor }}
--- a/packages/map-template/src/components/ChatButton/ChatButton.jsx
+++ b/packages/map-template/src/components/ChatButton/ChatButton.jsx
@@ -1,4 +1,5 @@
 import { createPortal } from 'react-dom';
+import i18n from 'i18next';
 import { useRecoilValue } from 'recoil';
 import PropTypes from 'prop-types';
 import chatHistoryState from '../../atoms/chatHistoryState';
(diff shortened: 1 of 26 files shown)


packages/map-template/src/components/LegendDialog/LegendDialog.jsx
```diff
@@ -57,7 +57,14 @@ function LegendDialog() {
...
         <div className="legend__background"></div>
-        <div className="legend" ref={legendModalRef}>
+        <div
+            className="legend"
+            ref={legendModalRef}
+            role="dialog"
+            aria-modal="true"
+            aria-labelledby="legend-dialog-title"
+        >
+            <h2 id="legend-dialog-title" className="legend__title">{t('Legend')}</h2>
```
vishak-mapspeople, line 67: - Untranslated: there is no 'Legend' key in any of the 9 locale files (verified — en.js has no capital-L token; de.js has only 'Show legend'). i18next falls back to the key, so the heading reads literal English "Legend" in every language — the exact thing this PR set out to fix.
- Unstyled + visible: legend__title appears in no SCSS (verified), so it renders as a browser-default h2 (large, bold, default margins) — a visible, unreviewed change to the legend dialog that can push/overflow content.

packages/map-template/src/components/ChatButton/ChatButton.jsx
```diff
@@ -1,4 +1,5 @@
 import { createPortal } from 'react-dom';
+import i18n from 'i18next';
```
vishak-mapspeople, line 2: Please address the coderabbit comment

vishak-mapspeople (approved):
Apart from the 2 suggestions, LGTM 👍

packages/map-template/src/components/ChatButton/ChatButton.jsx
Lucaci-Andrei: addressing the comment breaks the web app due to how we render the component 😄

PR #716, opened 2026-07-14
fix: filter even listener leak

Lucaci-Andrei:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where filter/highlight could repeatedly re-apply on map `idle`, leading to sustained CPU/GPU usage and frequent highlight wiping.
  * Updated map filter/highlight synchronization to apply immediately once the map instance is ready, matching the current floor right after navigation.
  * Changed map idle handling to run once for the initial positioning transition, preventing unnecessary render loops and unintended reset behavior.


vishak-mapspeople (approved):
LGTM 👍

PR #717, opened 2026-07-22
fix(a11y): add semantic h2 headings to major panels

Lucaci-Andrei:
## Summary
- Adds semantic `h2` headings to Search results, Directions, Wayfinding, Location Details, and Usage Consent so screen readers can navigate panel structure
- Visually hides Search/Directions headings where a visible title already exists; styles Location Details / Wayfinding headings to match existing look
- Adds `Search results` i18n keys across all 9 languages
- Skips LegendDialog (already has an `h2` from the aria-labels work)

Extracted from `feat/improve-accesibility-v2` as the next slice after `feat/extract-aria-labels`.

## Test plan
- [ ] Open Search results → confirm visually unchanged; VoiceOver/NVDA rotor shows an `h2` for search results
- [ ] Open Wayfinding → title still looks the same, announced as a heading
- [ ] Open Location Details → location name still looks the same, announced as a heading
- [ ] Start Directions → visually unchanged; heading announced as “Directions”
- [ ] Open chat Usage Consent → title level is `h2` (was `h3`)
- [ ] Switch language and confirm Search results heading translates

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Improved screen reader navigation by adding/adjusting semantic `h2` headings in search results, directions, wayfinding, location details, and usage consent panels.
  * Updated screen-reader-only heading styling to use modern `clip-path` via a shared `sr-only` mixin.
* **Localization**
  * Added the translated “Search results” label across supported languages.
* **Documentation**
  * Updated the unreleased changelog to reflect the accessibility and screen-reader-only styling improvements.

### DIFF
--- a/packages/map-template/src/components/Search/Search.scss
+++ b/packages/map-template/src/components/Search/Search.scss
@@ -107,4 +107,16 @@
             display: none;
         }
     }
+
+    &__sr-heading {
+        position: absolute;
+        width: 1px;
+        height: 1px;
+        padding: 0;
+        margin: -1px;
+        overflow: hidden;
+        clip: rect(0, 0, 0, 0);
+        white-space: nowrap;
+        border: 0;
+    }
 }
--- a/packages/map-template/src/components/Search/Search.scss
+++ b/packages/map-template/src/components/Search/Search.scss
@@ -109,14 +109,6 @@
     }
 
     &__sr-heading {
-        position: absolute;
-        width: 1px;
-        height: 1px;
-        padding: 0;
-        margin: -1px;
-        overflow: hidden;
-        clip: rect(0, 0, 0, 0);
-        white-space: nowrap;
-        border: 0;
+        @include variables.sr-only;
     }
 }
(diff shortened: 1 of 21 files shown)


packages/map-template/src/components/Search/Search.scss
```diff
@@ -107,4 +107,16 @@
...
+        position: absolute;
+        width: 1px;
+        height: 1px;
+        padding: 0;
+        margin: -1px;
+        overflow: hidden;
+        clip: rect(0, 0, 0, 0);
+        white-space: nowrap;
+        border: 0;
+    }
```
vishak-mapspeople, line 113: if you can put that block in one shared place (a reusable snippet SCSS calls a "mixin" or "placeholder", basically a named recipe you can pull in anywhere) and have both files point at it. Directions.scss and here

vishak-mapspeople (approved):
LGTM 👍

packages/map-template/src/components/Search/Search.scss
Lucaci-Andrei: Fixed 👍

PR #718, opened 2026-07-23
Fix form and live regions

Lucaci-Andrei:


vishak-mapspeople (approved):
Done well: the min-inline-size catch, reusing the existing sr-only mixin instead of re-inventing it, tight scope, and complete i18n.

PR #719, opened 2026-07-23
Feat/add include value in filter for components

Lucaci-Andrei:
## Summary by CodeRabbit

* **New Features**
  * Added an option to include each item’s value when filtering combo-box results.
  * Added configurable searchable text for combo-box items, allowing additional text to be matched during filtering.
  * Existing filtering behavior remains unchanged by default.

* **Documentation**
  * Updated combo-box and combo-box item documentation with the new filtering options.


packages/components/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [13.37.0] - 2026-07-21
```
vishak-mapspeople: changelog says minor release, label is patch
correct before release

packages/components/src/components/combo-box/combo-box.tsx
```diff
@@ -397,7 +406,13 @@ export class ComboBox {
     filter(): HTMLMiDropdownItemElement[] {
         if (this.filterElement) {
             const inputQuery: string = this.filterElement.value;
-            const miDropdownItemTexts: string[] = this.items.map(item => (item.text || item.innerText));
+            // When includeValueInFilter is set, append the item's `value` to the searchable text so
+            // callers can match on hidden data (e.g. an id) without it showing in the input/list.
+            const searchText = (item: HTMLMiDropdownItemElement): string => {
+                const text = item.text || item.innerText;
+                return this.includeValueInFilter ? `${text} ${item.value}` : text;
```
vishak-mapspeople: When an item has no value, item.value is undefined, so with the flag on this string becomes e.g. "Main Building undefined". The word "undefined" then becomes matchable — typing und/def/fin false-matches every value-less item, and typing "undefined" matches them all. Could we only append the value when it exists?
return this.includeValueInFilter && item.value ? `${text} ${item.value}` : text;
 (dropdown-item.tsx:12 declares @Prop() value: string; with no default, so it's undefined unless the caller sets it.)

vishak-mapspeople (approved):

matbmapspeople:
What Vishak said 😄 
LGTM otherwise.

packages/components/CHANGELOG.md
Lucaci-Andrei: Fixed 👍

packages/components/src/components/combo-box/combo-box.tsx
Lucaci-Andrei: Fixed 👍

PR #720, opened 2026-07-24
fix(a11y): enhance focus styles and improve accessibility

Lucaci-Andrei:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard focus indicators across chat controls for better accessibility.
  * Added reduced-motion support for users who prefer fewer animations.
  * Improved contrast for user chat messages.
  * Fixed responsive sizing for dialogs, panels, modals, and venue selectors.
  * Improved scrolling around sticky headers and enabled scrolling for lengthy content.
  * Resolved layout issues affecting location lists and wayfinding content.
  * Improved stacking behavior for sticky headers and notifications.

### DIFF
--- a/packages/map-template/src/components/ChatWindow/ChatInput/ChatInput.scss
+++ b/packages/map-template/src/components/ChatWindow/ChatInput/ChatInput.scss
@@ -49,13 +49,18 @@
     resize: none;
     min-height: 40px;
     max-height: 120px;
-    line-height: 1.4;
+    line-height: 1.5;
     background-color: transparent;
     margin: 0;
     vertical-align: middle;
 
     &:focus {
         outline: none;
+    }
+
+    &:focus-visible {
+        outline: 2px solid #0066cc;
+        outline-offset: 2px;
     }
 
     &:disabled {
@@ -90,6 +95,15 @@
         color: white;
     }
 
+    &:focus {
+        outline: none;
+    }
+
+    &:focus-visible {
+        outline: 2px solid #0066cc;
+        outline-offset: 2px;
+    }
+
     &:disabled {
         opacity: 0.5;
         cursor: not-allowed;
@@ -119,6 +133,15 @@
         opacity: 0.8;
     }
 
+    &:focus {
+        outline: none;
+    }
+
+    &:focus-visible {
+        outline: 2px solid #0066cc;
+        outline-offset: 2px;
+    }
+
     &:disabled {
         opacity: 0.5;
         cursor: not-allowed;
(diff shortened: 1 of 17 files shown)


packages/map-template/src/components/MapTemplate/MapTemplate.scss
```diff
@@ -1,6 +1,17 @@
 @import '@mapsindoors/midt/build/css/mapsindoors-map-variables.css';
 @import '@mapsindoors/css/toggle.css';
 
+@media (prefers-reduced-motion: reduce) {
```
vishak-mapspeople, line 4: This block targets * (every element) with !important and sits at the global root, unlike every other rule in this file which is scoped to .mapsindoors-map. Since map-template is embedded inside the CMS and customer sites, this reaches out and disables motion on the whole host page — and !important means the host can't override it. Could we scope it to our widget?
 @media (prefers-reduced-motion: reduce) {
   .mapsindoors-map *, .mapsindoors-map *::before, .mapsindoors-map *::after { … }
}

packages/map-template/src/components/ChatWindow/ChatInput/ChatInput.scss
```diff
@@ -58,6 +58,11 @@
         outline: none;
     }
 
+    &:focus-visible {
+        outline: 2px solid #0066cc;
```
vishak-mapspeople: The focus colour #0066cc is hardcoded here and repeated in four spots, so it won't follow a customer's theme and can drift if changed later. Worth hoisting to one variable/token. Also worth noting: the neighbouring ChatButton.scss:30 uses var(--chat-button-primary-color, #007bff) — a themeable variable and a different blue — so two adjacent chat controls currently show different focus rings. Could we make them consistent (same token, same colour)?

packages/map-template/src/components/Directions/Directions.scss
```diff
@@ -151,6 +151,7 @@
         overflow-y: auto;
         -webkit-overflow-scrolling: touch;
         touch-action: pan-y;
+        scroll-padding-top: 56px;
```
vishak-mapspeople: 56px is the sticky-header height assumed as a literal in four files. If a header height ever changes, focused items get mis-offset in each panel and there are four places to fix. Could this be one shared variable? Non-blocking.

Lucaci-Andrei: Fixed 👍

packages/map-template/src/components/MapTemplate/MapTemplate.scss
Lucaci-Andrei: Fixed 👍

packages/map-template/src/components/ChatWindow/ChatInput/ChatInput.scss
Lucaci-Andrei: Fixed 👍

PR #722, opened 2026-07-29
fix: add lazy loading for chat

Lucaci-Andrei:
Lazy-load the chat feature and remove stale build artifacts, reducing the npm package size from ~5 MB to ~3 MB.
Added loading spinner as fallback

## Summary by CodeRabbit

* **New Features**
  * Chat content now loads only when opened, with a clear loading indicator.
  * Map and chat loading failures display localized error messages with reload and close options.
  * Added improved fallback layouts for mobile and desktop experiences.
  * Added loading and error messages across supported languages.

* **Bug Fixes**
  * Prevented chat or map loading failures from disrupting surrounding interface content.
  * Corrected Spanish and Dutch translations for the “Close” action.


packages/map-template/src/components/Sidebar/Sidebar.jsx
```diff
@@ -150,12 +152,14 @@ function Sidebar({ directionsFromLocation, directionsToLocation, pushAppView, cu
             />
         </Modal>,
         <Modal isOpen={currentAppView === appViews.CHAT} key="CHAT">
-            <ChatWindow
-                isVisible={currentAppView === appViews.CHAT}
-                onClose={() => pushAppView(appViews.SEARCH)}
-                onSearchResults={handleChatLocations}
-                onShowRoute={handleChatShowRoute}
-            />
+            {currentAppView === appViews.CHAT && <Suspense fallback={<ChatLoadingFallback />}>
```
vishak-mapspeople: Not blocking this PR — more a recommendation for a follow-up. While reviewing I checked the whole monorepo and there's no ErrorBoundary anywhere (no componentDidCatch/getDerivedStateFromError, none named ErrorBoundary, react-error-boundary not installed). That's fine today because the only other lazy chunks — the Mapbox/Google views in MIMap.jsx — load instantly at mount, so a failed chunk almost never happens in practice.

Chat is different: it loads on demand, potentially long after the page opened, which is exactly the window where a redeploy can invalidate the chunk name a stale tab is holding (ChunkLoadError). Your Suspense fallback handles the waiting, but not the failing — if the chunk 404s or the network drops, React.lazy throws, and with no boundary that currently unmounts the whole widget, not just chat.

Since this is really a repo-wide gap (the map engines have it too), I'd suggest a small dedicated PR: one shared ErrorBoundary — ideally treating ChunkLoadError specially with a one-time location.reload() so a redeploy self-heals — wrapping the two map-engine Suspenses and the new chat one. Cheap, low-risk, and it hardens all three at once. Totally fine to merge this PR as-is and do that separately; flagging so it's on the radar rather than discovered via a "the map vanished when I clicked chat" ticket.

packages/map-template/src/components/ChatWindow/ChatLoadingFallback.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.chat-loading-fallback {
+    display: grid;
+    place-items: center;
+    min-height: 200px;
+
+    &__spinner {
+        width: var(--spacing-xxx-large);
+        height: var(--spacing-xxx-large);
+        border: 8px solid #00565533;
```
vishak-mapspeople: The spinner's track is a hardcoded #00565533, while the moving arc on the next line uses var(--chat-window-primary-color, #005655). So a customer who themes the chat colour gets a themed arc spinning on an un-themed default-teal track. Could the track derive from the same variable, e.g. color-mix(in srgb, var(--chat-window-primary-color, #005655) 20%, transparent)?

Lucaci-Andrei: Fixed 👍

packages/map-template/src/components/Sidebar/Sidebar.jsx
Lucaci-Andrei: Good catch, I will include an error boundary in this PR

Lucaci-Andrei: Error boundary added for lazy loaded components (Chat, Mapbox and GoogleMaps) ready for round 2 of reviews 😄

vishak-mapspeople (approved):

vishak-mapspeople (approved):
LGTM 👍

PR #727, opened 2026-08-13
fix: guard Modal MutationObserver against a nulled ref after unmount

Lucaci-Andrei:


packages/map-template/src/components/Sidebar/Modal/Modal.jsx
```diff
@@ -33,8 +33,9 @@ function Modal({ children, isOpen }) {
      * If the height of the content is bigger than the height of the modal, the fullHeight should be set.
      */
     useEffect(() => {
-        if (!contentRef) return;
```
matbmapspeople, line 36: Is there a need for 2 the same early exits?

Lucaci-Andrei: yes, the second one is in case the mutation observer fails to find the new layout, since we are switching from Mobile/Desktop

matbmapspeople (approved):

